PR #3, opened 2021-03-12
Feature/add swagger details

cookel2:
# What has changed

The /hello endpoint details have been removed from the swagger spec.

Details of the following future endpoints have been added to the swagger:

GET /jobs/{id}
- Request path parameter: {id} = id of a search reindex job as returned when job POSTed (Job.id)
- Request body: none
- Response body: Model - “Job”

GET /jobs
- Request body: none
- Request parameters:
    - Limit (see API Standards documentation)
    - Offset (see API Standards documentation)
    - Sort (sort on last_updated field)
    - State (filter option to bring back specific jobs on state)
- Response body: Model - “Jobs”

POST /jobs
  - Request body: Model - “Job” without id and links as these are generated by the application
  - Response body: Model - “Job”

Models
* = mandatory fields

Job resource:
{
  "id" *: string,
  "last_updated" *: ISODate, //ISO8601 timestamp
  "links" *: {
    "tasks" *: string // format: http://localhost:<port>/jobs/{id}/tasks
    "self" *: string // format: http://localhost:<port>/jobs/{id}
  },
  "number_of_tasks" : integer,
  "reindex_completed": ISODATE, //ISO8601 timestamp,
  "reindex_failed": ISODATE, //ISO8601 timestamp
  "reindex_started": ISODATE, //ISO8601 timestamp
  "search_index_name" *: string,
  "state" *: string, // enum [ created, in-progress, completed, failed ]
  "total_search_documents" : integer,
  "total_inserted_search_documents" : integer
}

Jobs resource:
{
  "count" *: integer
  "items" *: [
    <job>
  ],
  "limit" *: integer,
  "offset" *: integer,
  "total_count" *: integer,
  "sort": integer
}

# How to review

Check that the details in the swagger spec match the details in the specification above. 

Use the Swagger Editor to check that there are no errors: https://editor.swagger.io/

# Who can review

@nshumoogum, @jessjenkins, or anyone but me


swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
+  id:
+    name: id
+    description: The id of a search reindex job when job POSTed (Job.id)
+    in: path
+    required: true
+    type: string
+
+  limit:
+    name: limit
+    description: The maximum number of Job objects we're returning in this Jobs object
```
sandypadmanabhan, line 199: the default value is 20 and the maximum allowed is 1000( I think)

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
+    in: query
+    type: integer
```
sandypadmanabhan, line 201: default: 20 value can be added to the list

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
+
+  offset:
+    name: offset
+    description: The number of Job objects into the full list that this particular response is starting at
+    in: query
+    type: integer
```
sandypadmanabhan, line 207: Do we have a default for offset?

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
     url: "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"
+
 basePath: "/v1"
+
 schemes:
   - http
+
 tags:
-  - name: "hello"
+  - name: "search-reindex"
```
sandypadmanabhan, line 17: usually this is either "public", private
```
tags:
- name: "Public"
- name: "Private"
```

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
+
 paths:
-  /hello:
+
+  /jobs/{id}:
+    get:
+      description: Get a specific job.
```
sandypadmanabhan, line 23: Tags can be added to get
  get:
      tags:
      - "Public"

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
+        500:
+          $ref: '#/responses/InternalError'
+
+  /jobs:
     get:
-      tags:
-        - hello
-      summary: Example hello world endpoint
-      description: Returns an example payload.
+      description: Get a list of jobs.
```
sandypadmanabhan, line 38: tags to be added

swagger.yaml
```diff
@@ -1,42 +1,233 @@
...
       responses:
         200:
           description: OK
           schema:
-            $ref: "#/definitions/HelloResponse"
+            $ref: "#/definitions/Jobs"
+        500:
+          $ref: '#/responses/InternalError'
+    post:
+      description: Post new job details and receive the new Job object in response.
```
sandypadmanabhan, line 54: We usually add a summary before a description
 ```
    tags:
        - "Public"
      summary: "Returns the latest version of an area"
      description: "Returns an area for a given id"
```

File: swagger.yaml (hunk 1 above)
Comment (cookel2) [reply swagger.yaml:199]: ok

File: swagger.yaml (hunk 2 above)
Comment (cookel2) [reply swagger.yaml:201]: will do

Comment (cookel2) [reply swagger.yaml:210]: done

File: swagger.yaml (hunk 3 above)
Comment (cookel2) [reply swagger.yaml:217]: yes, done

File: swagger.yaml (hunk 4 above)
Comment (cookel2) [reply swagger.yaml]: done

File: swagger.yaml (hunk 5 above)
Comment (cookel2) [reply swagger.yaml]: done

File: swagger.yaml (hunk 6 above)
Comment (cookel2) [reply swagger.yaml]: done

File: swagger.yaml (hunk 7 above)
Comment (cookel2) [reply swagger.yaml]: now added summaries to all three endpoints

File: swagger.yaml [hunk 8]
```diff
@@ -1,42 +1,244 @@
...
+        - "Public"
+      summary: Get a specific job.
+      description: Get the specific search reindex job that has the id given in the path.
+      parameters:
+        - $ref: '#/parameters/id'
       produces:
         - application/json
       responses:
         200:
           description: OK
```
sandypadmanabhan, line 33: It would be better to say something along the lines of "Successfully returned a job for a given id" instead of just an OK

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
           schema:
-            $ref: "#/definitions/HelloResponse"
+            $ref: "#/definitions/Job"
```
sandypadmanabhan, line 35: We might want to add a 404 to this too when the resource is not found

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      parameters:
+        - $ref: '#/parameters/limit'
+        - $ref: '#/parameters/offset'
+        - $ref: '#/parameters/sort'
+        - $ref: '#/parameters/state'
+      produces:
+        - application/json
+      responses:
+        200:
+          description: OK
```
sandypadmanabhan, line 54: similar comment as above

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+          schema:
+            $ref: "#/definitions/Jobs"
```
sandypadmanabhan, line 56: As it returns a json object containing a list of jobs, Should it be returning the properties such as count,items, limit, offset, total count ,state etc?
An error 400 for invalid request can be added to the response as well

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+        - "Public"
+      summary: Create a new search reindex job.
+      description: Post details of a new search reindex job and receive the new Job object created in response.
+      parameters:
+        - $ref: '#/parameters/job'
+      produces:
+        - application/json
+      responses:
+        200:
+          description: OK
```
sandypadmanabhan, line 70: A brief description would be good

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+          schema:
+            $ref: "#/definitions/Job"
         500:
```
sandypadmanabhan, line 73: It could have other errors returned, such as invalid request

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      - "last_updated"
+      - "links"
+      - "search_index_name"
+      - "state"
+    properties:
+      id:
+        type: string
+      last_updated:
+        type: string
+        format: date
```
sandypadmanabhan, line 96: Should this be date-time, if it has a timestamp?

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+            format: http://localhost:<port>/jobs/{id}/tasks
+          self:
+            type: string
+            format: http://localhost:<port>/jobs/{id}
+      number_of_tasks:
+        type: integer
+      reindex_completed:
+        type: string
+        format: date
+        description: ISO8601 timestamp
```
sandypadmanabhan, line 115: A brief description would be helpful. eg Something along the lines of "The date when the reindex was completed"

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      reindex_failed:
+        type: string
+        format: date
+        description: ISO8601 timestamp
```
sandypadmanabhan, line 119: Same as above

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      reindex_started:
+        type: string
+        format: date
+        description: ISO8601 timestamp
```
sandypadmanabhan, line 123: Same as above

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      - "links"
+      - "search_index_name"
+      - "state"
+    properties:
+      id:
+        type: string
+      last_updated:
+        type: string
+        format: date
+        description: ISO8601 timestamp
```
sandypadmanabhan, line 97: A brief description would be helpful

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      links:
+        type: object
+        required:
+          - "tasks"
+          - "self"
+        properties:
+          tasks:
+            type: string
+            format: http://localhost:<port>/jobs/{id}/tasks
```
sandypadmanabhan, line 106: An example would be helpful

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+          self:
+            type: string
+            format: http://localhost:<port>/jobs/{id}
```
sandypadmanabhan, line 109: An example value would be helpful

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+      total_count:
+        type: integer
+      sort:
+        type: integer
+
+parameters:
+
+  id:
+    name: id
+    description: The id of a search reindex job when job POSTed (Job.id)
```
sandypadmanabhan, line 201: It is better to keep this generic as it is used during GET as well

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+        type: integer
+
+  Jobs:
+    type: object
+    required:
+      - "count"
+      - "items"
+      - "limit"
+      - "offset"
+      - "total_count"
```
sandypadmanabhan, line 180: Not sure if total count and items are required when a request is made. I know limit , offset and count will be passed as query params, but not sure about total count and items

swagger.yaml
```diff
@@ -1,42 +1,244 @@
...
+    in: path
+    required: true
+    type: string
+
+  limit:
+    name: limit
+    description: The maximum number of Job objects we're returning in this Jobs object. The default limit is 20.
+    in: query
+    type: integer
+    default: 20
```
sandypadmanabhan, line 211: required : true  to be added to limit and offset as they are needed when requesting list of jobs. Do we need sort and state to be added to the query parameter? if so they need this field applied as well

sandypadmanabhan (changes requested):
Requested changes

swagger.yaml
nshumoogum: I concur

swagger.yaml
nshumoogum: I concur

swagger.yaml
nshumoogum: Initially the request to get list of jobs actually will not result in any 400 requests (no path, query parameters or a  body to this endpoint) at this point in time.

Future work (stories) will add pagination which could result in 400, but we will add in these details when we come to it. Thanks Sandy :)

swagger.yaml
nshumoogum: Yes it should be 201 when creating a resource and concur a more detailed description is helpful here

swagger.yaml
nshumoogum: Again there will be no path or query parameters, nor a body so for now I cannot perceive any other errors besides internal server errors.

swagger.yaml
nshumoogum: Yes this should be a `date-time` format, worthwhile adding an example of ISO8601 format:

`example: "2020-06-11T11:49:21.520922Z"`

swagger.yaml
nshumoogum: Yes a description to explain what this field represents, something like:

`A ISO8601 timestamp representing the last time the resource was updated.`

swagger.yaml
cookel2: done

swagger.yaml
cookel2: that makes sense - done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: ok, done

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
+      reindex_started:
+        type: string
+        format: date
+        description: ISO8601 timestamp
+      search_index_name:
+        type: string
+      state:
+        type: string
+        enum:
+          - CREATED
```
nshumoogum, line 139: Values should be lower-kebab-case

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
+          - IN-PROGRESS
+          - COMPLETED
+          - FAILED
+      total_search_documents:
```
nshumoogum, line 143: Every field should contain a description. This should be very explicit, see other swagger specs such as the [dataset API](https://github.com/ONSdigital/dp-dataset-api/blob/develop/swagger.yaml) or [recipe API](https://github.com/ONSdigital/dp-recipe-api/blob/develop/swagger.yaml).

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
+
+  NewJob:
     type: object
+    required:
+      - "last_updated"
+      - "search_index_name"
+      - "state"
     properties:
-      message:
+      last_updated:
```
nshumoogum, line 153: All fields containing ISO8601 timestamps should be of format `date-time`, nice to always include an example and explicit descriptions for each field.

Incl. `reindex_started`, `reindex-completed` and `reindex-failed`

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
-        description: "Message returned by hello world endpoint"
-        example: "Hello, world!"
\ No newline at end of file
+        format: date
+        description: ISO8601 timestamp
+      search_index_name:
+        type: string
+      state:
+        type: string
+        enum:
```
nshumoogum, line 175: See enum comment above, lower-kebab-case

swagger.yaml
nshumoogum: I concur, so this will be the unique identifier for a search reindex job resource

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
+  id:
+    name: id
+    description: The id of a search reindex job when job POSTed (Job.id)
+    in: path
+    required: true
+    type: string
+
+  limit:
+    name: limit
+    description: The maximum number of Job objects we're returning in this Jobs object. The default limit is 20.
```
nshumoogum, line 218: I think it is better to refer to a job as a `resource` instead of an `object`

swagger.yaml
nshumoogum: limit, offset, sort and state are all optional fields; so they should **not** be set to required. Values offset, limit and sort will have defaults.

state query parameter will need to handle a list of comma separated values and are restricted to a list of enums. This parameter is also not required as the API uses this parameter to filter by state(s)

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
+  offset:
+    name: offset
+    description: The number of Job objects into the full list that this particular response is starting at. The default number to start at is 0.
+    in: query
+    type: integer
+    default: 0
+
+  sort:
+    name: sort
+    description: Sort on last_updated field
```
nshumoogum, line 232: The default for sort parameter will be on the `last_updated` field, so can add default parameter to sort.

The sort query parameter could be set to something other than `last_updated` and hence this must be clear in the description, something like:

`description: The sort order in which to return a list of reindex job resources. This defaults to last_updated in ascending order.`

swagger.yaml
```diff
@@ -1,42 +1,254 @@
...
     url: "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"
+
 basePath: "/v1"
+
 schemes:
   - http
+
 tags:
-  - name: "hello"
+  - name: "Public"
```
nshumoogum, line 17: This api will never be made public. All endpoints are for internal use only.

So we have `"Private User"` and `"Private"` endpoints only. All the ones currently in the swagger spec should reference the `"Private User"`

swagger.yaml
nshumoogum: After reviewing the rest of the swagger spec, I see we have added query parameters for limit, offset, sort and state so Sandy is right there will be the possibility of 400 bad request.

nshumoogum (changes requested):

swagger.yaml
nshumoogum: Future endpoints will be for `"Private"` use only. This means backend applications are the only apps that will make use of these endpoints and not users (e.g. Florence - internal CMS)

swagger.yaml
nshumoogum: The reason for being explicit is that it helps explain what the fields are, usually we use the swagger yaml file to generate documentation for our APIs on the developer site. Due to this application being solely private it wont be publicised in the same manner as some of our other APIs.

swagger.yaml
cookel2: accepted, changed

swagger.yaml
cookel2: invalid request added

swagger.yaml
cookel2: that makes sense - done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: example added

swagger.yaml
cookel2: example added

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done - I had wondered what case I should use so that's good to know

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: example added

swagger.yaml
cookel2: example added - good idea

swagger.yaml
cookel2: done

swagger.yaml
cookel2: I'm not sure if I've done this correctly. The Reindex API spec shows * next to each of the fields in the 'Jobs' resource, except for the sort field, and so I understood this to mean that those fields are required. However, I've made the swagger now show them as optional.

swagger.yaml
cookel2: descriptions now added - it would be good to know how to generate documentation from the swagger yaml

swagger.yaml
cookel2: examples and descriptions now added but I'm not sure if the examples I've used are correct?

swagger.yaml
cookel2: done

swagger.yaml
cookel2: done

swagger.yaml
cookel2: The sort field is confusing me a little. At first I took it to be a boolean i.e. true means sort on the last_updated field, false means don't sort. However, I noticed that the spec has it down as an integer. In that case I'm not sure what its default value should be. However, the way you have described it here makes it sound like a string with a default value of 'last_updated' so that's what I've done. Is that correct?

swagger.yaml
cookel2: replaced all instances of "Public" with "Private User' now

swagger.yaml
nshumoogum: So these fields are optional in the request (query parameters) but required/mandatory in the response body.

swagger.yaml
cookel2: I've found the following example of an ISO8601 timestamp so I've replaced my previous examples with this: 2021-03-18T09:03:41+0000

swagger.yaml
cookel2: Nathan has confirmed that sort is a string

swagger.yaml
```diff
@@ -104,7 +104,7 @@ definitions:
         type: string
         format: date-time
         description: "A ISO8601 timestamp representing the last time the resource was updated."
-        example: "2020-04-26T09:51:03-00:00"
+        example: "2021-03-18T09:03:41+0000"
```
nshumoogum, line 107: Good spot

nshumoogum (dismissed):

swagger.yaml
cookel2: thanks Nathan

nshumoogum (dismissed):

swagger.yaml
```diff
@@ -1,42 +1,223 @@
...
+        description: "The total number of documents that are held within the search index in question."
+      total_inserted_search_documents:
+        type: integer
+        description: "The total number of documents inserted into the new search index for a reindex job."
+
+  Jobs:
     type: object
+    description: "A list of Job resources held in an array."
+    required:
+      - "count"
```
sandypadmanabhan, line 164: Discussed with Nathan and -limit, offset, sort  is mandatory , so it needs to be added to the list.

cookel2: ok, done

swagger.yaml
```diff
@@ -1,42 +1,223 @@
...
+      limit:
+        type: integer
+        description: "The max number of Job resources we're returning in this response. The default is 20."
+      offset:
+        type: integer
+        description: "The number of Job resources into the full list that this particular response is starting at. The default is 0."
+      total_count:
+        type: integer
+        description: "How many total items there may be (so the full list size, maybe thousands)."
+      sort:
```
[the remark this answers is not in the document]
cookel2: state added but please check that I've given it the correct type and description (as I'm not sure if I have)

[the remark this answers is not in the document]
sandypadmanabhan: yes, the description is similar to line 146 ` The current state of this search reindex job`. And please can you add examples too please, for each field limit, count, sort etc.(if that's ok). This gives us a clear model to look at and will be easy for us to understanc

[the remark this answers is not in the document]
cookel2: ok, done

swagger.yaml
```diff
@@ -1,42 +1,236 @@
...
         type: string
-        description: "Message returned by hello world endpoint"
-        example: "Hello, world!"
\ No newline at end of file
+        description: "The sort order of the list."
+        example: number_of_tasks
+      state:
+        type: string
+        description: "The current state(s) of the Job resources in the list."
+        example: "created, failed"
```
sandypadmanabhan, line 199: just "created" should work

sandypadmanabhan (approved):

PR #4, opened 2021-03-29
Feature/add post jobs endpoint

cookel2:
# What has changed

An endpoint has been added that allows a POST request to be made, which results in the creation of a Job resource containing default values, as follows:

{
    "id": "331f6b96-fcd5-4c52-b77d-8622162b7510", NB. This is a random UUID
    "last_updated": "2021-03-29T08:22:15.593168Z", NB. This defaults to the current time
    "links": {
        "tasks": "http://localhost:12150/jobs/331f6b96-fcd5-4c52-b77d-8622162b7510/tasks",
        "self": "http://localhost:12150/jobs/331f6b96-fcd5-4c52-b77d-8622162b7510"
    },
    "number_of_tasks": 0,
    "reindex_completed": "0001-01-01T00:00:00Z",
    "reindex_failed": "0001-01-01T00:00:00Z",
    "reindex_started": "0001-01-01T00:00:00Z",
    "search_index_name": "Default Search Index Name",
    "state": "created",
    "total_search_documents": 0,
    "total_inserted_search_documents": 0
}

The Job resource is then stored in a map, which uses the id value as the key value. Logging has been added to show the details of the created Job and the length of the map (after thee Job has been added to it).

Unit tests have been added for the api, the handler, and the storer.

A component test has also been added, which tests that the expected default values are returned.

# How to review

The endpoint can be tested locally using this POST URL: http://localhost:25700/jobs

The component test can be run using this command: go test -component

# Who can review

Anyone but me


api/jobs.go
```diff
@@ -0,0 +1,47 @@
...
+			http.Error(w, "Failed to write http response", http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+//NewID generates a random uuid and returns it as a string.
+var NewID = func() string {
+	return uuid.NewV4().String()
+}
```
daiLlew, line 47: Can you move this var to the top of the file between imports and the function please? General convention for file structure imports, const + vars, funcs 😄

api/jobs_test.go
```diff
@@ -0,0 +1,79 @@
...
+				So(newJob.ID, ShouldEqual, expectedJob.ID)
+				So(newJob.Links, ShouldResemble, expectedJob.Links)
+				So(newJob.NumberOfTasks, ShouldEqual, expectedJob.NumberOfTasks)
+				So(newJob.ReindexCompleted, ShouldEqual, expectedJob.ReindexCompleted)
+				So(newJob.ReindexFailed, ShouldEqual, expectedJob.ReindexFailed)
+				So(newJob.ReindexStarted, ShouldEqual, expectedJob.ReindexStarted)
+				So(newJob.SearchIndexName, ShouldEqual, expectedJob.SearchIndexName)
+				So(newJob.State, ShouldEqual, expectedJob.State)
+				So(newJob.TotalSearchDocuments, ShouldEqual, expectedJob.TotalSearchDocuments)
+				So(newJob.TotalInsertedSearchDocuments, ShouldEqual, expectedJob.TotalInsertedSearchDocuments)
```
daiLlew, line 56: Consider putting this inside its own convey statement.

features/steps/jobs_steps.go
```diff
@@ -0,0 +1,154 @@
...
+	uuid "github.com/satori/go.uuid"
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type JobsFeature struct {
```
daiLlew, line 25: Godoc.

features/steps/jobs_steps.go
```diff
@@ -0,0 +1,154 @@
...
+	svc            *service.Service
+	errorChan      chan error
+	Config         *config.Config
+	HTTPServer     *http.Server
+	ServiceRunning bool
+	ApiFeature     *componenttest.APIFeature
+	responseBody   []byte
+}
+
+func NewJobsFeature() (*JobsFeature, error) {
```
daiLlew, line 36: Godoc

features/steps/jobs_steps.go
```diff
@@ -0,0 +1,154 @@
...
+	f := &JobsFeature{
+		HTTPServer:     &http.Server{},
+		errorChan:      make(chan error),
```
daiLlew, line 39: does this need a buffer size?

features/steps/jobs_steps.go
```diff
@@ -0,0 +1,154 @@
...
+	return &hc, nil
+}
+
+func (f *JobsFeature) iWouldExpectIdLast_updatedAndLinksToHaveThisStructure(table *godog.Table) error {
+	f.responseBody, _ = ioutil.ReadAll(f.ApiFeature.HttpResponse.Body)
+	assist := assistdog.NewDefault()
+
+	expectedResult, err := assist.ParseMap(table)
+	if err != nil {
+		panic(err)
```
daiLlew, line 103: Is this panic intentional or should it return the error?

models/job.go
```diff
@@ -0,0 +1,25 @@
...
+	ReindexCompleted             time.Time `json:"reindex_completed"`
+	ReindexFailed                time.Time `json:"reindex_failed"`
+	ReindexStarted               time.Time `json:"reindex_started"`
+	SearchIndexName              string    `json:"search_index_name"`
+	State                        string    `json:"state"`
+	TotalSearchDocuments         int       `json:"total_search_documents"`
+	TotalInsertedSearchDocuments int       `json:"total_inserted_search_documents"`
+}
+
+type JobLinks struct {
```
daiLlew, line 22: Godoc

store/job_storer.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/ONSdigital/log.go/log"
+	"time"
+)
+
+type JobStorer struct {
+	JobsMap map[string]models.Job
+}
+
+// CreateJob creates a new Job resource and stores it in the JobsMap
+func (js *JobStorer) CreateJob(ctx context.Context, id string) (models.Job, error) {
```
daiLlew, line 15: Do we need to validate the ID? I.e. check if its empty, is it a duplicate etc.

store/job_storer.go
```diff
@@ -0,0 +1,43 @@
...
+		ID:          id,
+		LastUpdated: time.Now().UTC(),
+		Links: &models.JobLinks{
+			Tasks: "http://localhost:12150/jobs/" + id + "/tasks",
+			Self:  "http://localhost:12150/jobs/" + id,
+		},
+		NumberOfTasks:                0,
+		ReindexCompleted:             time.Time{}.UTC(),
+		ReindexFailed:                time.Time{}.UTC(),
+		ReindexStarted:               time.Time{}.UTC(),
```
daiLlew, line 28: Use  `time.Now().UTC()` instead.

Each of these time fields will be different (admitted only by nanoseconds) - I'm not sure if this matters or not but consider creating a single time var before you create the struct and using the var for each of the time fields.

store/job_storer.go
```diff
@@ -0,0 +1,43 @@
...
+		State:                        "created",
+		TotalSearchDocuments:         0,
+		TotalInsertedSearchDocuments: 0,
+	}
+	//Only create a new JobsMap if one does not exist already
+	if js.JobsMap == nil {
+		js.JobsMap = make(map[string]models.Job)
+	}
+
+	js.JobsMap[id] = newJob
```
daiLlew, line 39: Question: What happens if a job with this ID already exists? I don't know the acceptance criteria for this task but I'm assuming we shouldn't be able to create a new job with an existing ID? Depending on the requirements agreed for this task we should probably also capture the desired behaviour with some tests.

store/job_storer.go
```diff
@@ -0,0 +1,43 @@
+package store
+
+import (
+	"context"
+	models "github.com/ONSdigital/dp-search-reindex-api/models"
+	"github.com/ONSdigital/log.go/log"
+	"time"
+)
+
+type JobStorer struct {
```
daiLlew, line 10: Godoc.

store/job_storer.go
```diff
@@ -0,0 +1,43 @@
...
+	JobsMap map[string]models.Job
+}
+
+// CreateJob creates a new Job resource and stores it in the JobsMap
+func (js *JobStorer) CreateJob(ctx context.Context, id string) (models.Job, error) {
+
+	log.Event(ctx, "creating job", log.Data{"id": id})
+	newJob := models.Job{
```
daiLlew, line 18: Consider creating a `NewJob(id string)` func in the models package - it removes the burden on the caller to init a job object with the default values.

store/job_storer_test.go
```diff
@@ -0,0 +1,58 @@
...
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+// Constants for testing
+const (
+	testJobID1 = "UUID1"
+)
+
+var ctx = context.Background()
+
```
daiLlew, line 17: As already commented are you able to confirm what the expected/required behaviour is around attempting to create a new job with an existing ID?
Once that is clarified can you add some test cases to cover those requirements?

store/job_storer_test.go
```diff
@@ -0,0 +1,58 @@
...
+			So(job.ReindexStarted, ShouldEqual, expectedJob.ReindexStarted)
+			So(job.SearchIndexName, ShouldEqual, expectedJob.SearchIndexName)
+			So(job.State, ShouldEqual, expectedJob.State)
+			So(job.TotalSearchDocuments, ShouldEqual, expectedJob.TotalSearchDocuments)
+			So(job.TotalInsertedSearchDocuments, ShouldEqual, expectedJob.TotalInsertedSearchDocuments)
+		})
+	})
+}
+
+func expectedJob() models.Job {
```
daiLlew, line 41: 3rd time you've had to create a job with default values like this - consider making a `NewJob(id string)` func to do this for you.

daiLlew (changes requested):

api/jobs.go
cookel2: makes sense - done

api/jobs_test.go
cookel2: good idea - done

features/steps/jobs_steps.go
cookel2: added now

features/steps/jobs_steps.go
cookel2: I'm not sure. It works ok with the default of no buffer size but I'm happy to put a size in if there's one you would suggest.

store/job_storer.go
cookel2: The id gets generated by the service as a random uuid so in theory it should not try to create a new job with an existing ID. I'm not sure what would happen if it did but then I'm also not sure whether that scenario is something that needs to be handled (given that it's unlikely to happen). Unfortunately my tech lead is on leave this week, my guess would be that I should leave it as is, but what would you suggest?

store/job_storer.go
cookel2: Godoc now added

store/job_storer.go
cookel2: very good advice - done

store/job_storer_test.go
cookel2: please refer to earlier reply - I don't think these scenarios are likely to occur since the id gets generated by the service

store/job_storer_test.go
cookel2: done - I've now removed the expectedJob function altogether as the NewJob can be used instead

cookel2:
The reason why the audit tests have been failing appears to be as follows: the nancy command has found a vulnerability, which is an indirect dependency of something called scram. However, the latest version of scram has a new path in github, which also need to be changed (instead of http://github.com/xdg/scram it is now http://github.com/xdg-go/scram I believe). The problem is that if I manually amend this path and version in the sum.go file then it reverts back when I try to commit and push the code. To prevent this from happening I need to update the dependency that is actually using scram directly. That dependency is actually mongo-driver, which itself is a dependency in dp-component-test, which is a dependency in this project. But on further investigation, by Justin, it turns out that mongo-driver itself does not have the latest version of scram. So this build cannot be fixed until a new version of mongo-driver is released and then dp-component-test is updated with it. Here's the issue: https://jira.mongodb.org/projects/GODRIVER/issues/GODRIVER-1935?filter=allissues

store/job_storer.go
daiLlew: For me personally there is 2 main reasons why we should check the validity of the the ID & adding tests to cover those scenarios:

1 - Although the ID is generated by this service it's generated in another package. By assuming how the ID generation works and that it always gives us a valid unique value we are coupling packages and breaking encapsulation. We are relying on another package working correctly and not changing in future. As a general rule I personally think it's sensible to _never_ trust any input from an external source even if it is within this application. It is more work but I feel it's worth it because it reduces the chance of potential bugs.

2 - A duplicate ID has potentially destructive consequences - we could lose an existing job record if we were to overwrite it, or we could end up in an inconsistent state and I assume we don't want that. The complexity of implementing a check for a duplicate ID is reasonably small. I'd argue a small amount of extra work to be defensive and protect against this scenario is worth it.

As I said I appreciate requires more work but I personally feel it's worth it because it reduces the change of bugs and errors, provides better test coverage and gives us greater confidence in the behaviour of our apps.

Obviously it's up to you if you want to implement these changes or not. As a compromise do you think you could you ask one of the other tech leads for input - I'll go with whatever they decided.

api/jobs.go
```diff
@@ -0,0 +1,47 @@
...
+	log.Event(ctx, "Entering CreateJobHandler function, which generates a new Job resource.", log.INFO)
+	return func(w http.ResponseWriter, req *http.Request) {
+		ctx := req.Context()
+
+		id := NewID()
+
+		//Create job in job store
+		newJob, err := api.jobStore.CreateJob(ctx, id)
+		if err != nil {
+			log.Event(ctx, "creating and storing job failed", log.Error(err), log.ERROR)
```
daiLlew, line 27: Does this need to set an error `http.Status` on the response?

api/jobs_test.go
```diff
@@ -0,0 +1,82 @@
...
+					So(newJob.ReindexFailed, ShouldEqual, expectedJob.ReindexFailed)
+					So(newJob.ReindexStarted, ShouldEqual, expectedJob.ReindexStarted)
+					So(newJob.SearchIndexName, ShouldEqual, expectedJob.SearchIndexName)
+					So(newJob.State, ShouldEqual, expectedJob.State)
+					So(newJob.TotalSearchDocuments, ShouldEqual, expectedJob.TotalSearchDocuments)
+					So(newJob.TotalInsertedSearchDocuments, ShouldEqual, expectedJob.TotalInsertedSearchDocuments)
+				})
+			})
+		})
+	})
```
daiLlew, line 62: Consider adding another test to cover the scenario where `api.jobStore.CreateJob(ctx, id)` returns an error.

api/jobs_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func expectedJob() models.Job {
```
daiLlew, line 65: You can remove this and use the `New` func you added.

apierrors/errors.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"errors"
+)
+
+// A list of error messages for JobstorerAPI API
+var (
+	ErrJobNotFound         = errors.New("job not found")
+	ErrUnableToReadMessage = errors.New("failed to read message body")
+	ErrUnableToParseJSON   = errors.New("failed to parse json body")
```
daiLlew, line 11: Godoc.

features/steps/jobs_steps.go
daiLlew: I'm not sure tbh - if its working as expected I am happy with how it is.

store/job_storer.go
cookel2: For the sake of a good quality app I will go with what you suggest and I think you're right - it would be nice to be confident that the id won't be duplicated. In that case I will add validation and tests as requested.

models/job.go
```diff
@@ -0,0 +1,46 @@
...
+func NewJob(id string) Job {
+	return Job{
+		ID:          id,
+		LastUpdated: time.Now().UTC(),
+		Links: &JobLinks{
+			Tasks: "http://localhost:12150/jobs/" + id + "/tasks",
+			Self:  "http://localhost:12150/jobs/" + id,
+		},
+		NumberOfTasks:                0,
+		ReindexCompleted:             time.Time{}.UTC(),
```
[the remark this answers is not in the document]
cookel2: The time.Time{}.UTC() is used because it gives a zero time value, which is needed here rather than the current time. This is because when the reindex Job is first created then the 'ReindexCompleted' won't yet have a value - it won't have been completed yet.

features/steps/jobs_steps.go
cookel2: To be honest this code came from Armakuni so I don't know the answer. I can return the error if you think that would be better?

store/job_storer.go
cookel2: yes, agreed as discussed - done

store/job_storer.go
cookel2: time.Time{}.UTC() gives zero time (as the reindex is neither started, failed, nor completed when the job details are first created) but I have made a variable, as suggested, for clarity and to avoid repeating code

store/job_storer.go
cookel2: validation and tests now added

api/jobs.go
cookel2: yes it does, thanks, now added

api/jobs_test.go
cookel2: done

api/jobs_test.go
cookel2: good idea - done

apierrors/errors.go
cookel2: I've removed errors.go instead now as I wasn't using it.

models/job.go
```diff
@@ -0,0 +1,48 @@
...
+
+//JobLinks is a type that contains links to the endpoints for returning a specific job (self), and the tasks that it contains (tasks), respectively.
+type JobLinks struct {
+	Tasks string `json:"tasks"`
+	Self  string `json:"self"`
+}
+
+//NewJob returns a new Job resource that it creates and populates with default values.
+func NewJob(id string) Job {
+	zeroTime := time.Time{}.UTC()
```
daiLlew, line 30: Creating an empty `time.Time{}` will create a time instance with the _Zero value_ for time. I'm not sure if this is intentional or not? See t[his for explanation ](https://stackoverflow.com/questions/23051973/what-is-the-zero-value-for-time-time-in-go).

If this is the intended behaviour - mark this resolved, otherwise use `time.Now()` to create a new time instance.

features/steps/jobs_steps.go
daiLlew: I'm not sure either I've not see or worked with Armakuni yet or seen examples of how to write these tests so I cannot comment. If its taken from their example happy to leave as is.

daiLlew (approved):

PR #7, opened 2021-04-27
Feature/add get jobs endpoint

cookel2:
# What has changed

An endpoint has been added, which allows a GET request to be made, to provide details of all the search reindex jobs currently held in the job store. This results in the relevant Job resources being returned, in a list, or otherwise a null value is returned (if no Job resources exist).

NB. The swagger spec. is slightly different to the current functionality because the trello ticket specifies the following: ignore request query parameters, offset, limit, sort and state for now.

Example of the GET request:
http://localhost:25700/jobs

Example of a list of two Job resources being returned:
```json
{
    "jobs": [
        {
            "id": "fc0c8f28-0f2a-4ded-b27c-18f2d8748b76",
            "last_updated": "2021-04-27T14:27:13.599666Z",
            "links": {
                "tasks": "http://localhost:12150/jobs/fc0c8f28-0f2a-4ded-b27c-18f2d8748b76/tasks",
                "self": "http://localhost:12150/jobs/fc0c8f28-0f2a-4ded-b27c-18f2d8748b76"
            },
            "number_of_tasks": 0,
            "reindex_completed": "0001-01-01T00:00:00Z",
            "reindex_failed": "0001-01-01T00:00:00Z",
            "reindex_started": "0001-01-01T00:00:00Z",
            "search_index_name": "Default Search Index Name",
            "state": "created",
            "total_search_documents": 0,
            "total_inserted_search_documents": 0
        },
        {
            "id": "f207826a-e8ad-4202-a5f0-9f10e2a32e2c",
            "last_updated": "2021-04-27T14:27:15.50031Z",
            "links": {
                "tasks": "http://localhost:12150/jobs/f207826a-e8ad-4202-a5f0-9f10e2a32e2c/tasks",
                "self": "http://localhost:12150/jobs/f207826a-e8ad-4202-a5f0-9f10e2a32e2c"
            },
            "number_of_tasks": 0,
            "reindex_completed": "0001-01-01T00:00:00Z",
            "reindex_failed": "0001-01-01T00:00:00Z",
            "reindex_started": "0001-01-01T00:00:00Z",
            "search_index_name": "Default Search Index Name",
            "state": "created",
            "total_search_documents": 0,
            "total_inserted_search_documents": 0
        }
    ]
}
```

Example of a null value being returned if no Job resources exist:
```json
{
    "jobs": null
}
```

Logging has been added to show the sorted list of Job resources if returned successfully. 

Unit tests have been added for the api, the handler, and the storer.

A component test has also been added, which tests that the expected number and details of Job resources will be returned in chronological order of last_updated.

# How to review

The endpoint can be tested locally by first using this POST URL a number of times to create that number of Job resources in the Job Store: http://localhost:25700/jobs
The endpoint, which has a similar URL, can then be used as a GET URL to retrieve those same jobs in a single sorted list: http://localhost:25700/jobs

# Who can review

Anyone but me

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+
 	store "github.com/ONSdigital/dp-search-reindex-api/store"
 	"github.com/gorilla/mux"
 )
--- a/api/api.go
+++ b/api/api.go
@@ -21,5 +21,6 @@ func Setup(ctx context.Context, router *mux.Router, jobStore store.JobStore) *Jo
 
 	router.HandleFunc("/jobs", api.CreateJobHandler(ctx)).Methods("POST")
 	router.HandleFunc("/jobs/{id}", api.GetJobHandler(ctx)).Methods("GET")
+	//router.HandleFunc("/jobs", api.GetJobsHandler(ctx)).Methods("GET")
 	return api
 }
--- a/api/api.go
+++ b/api/api.go
@@ -22,6 +22,6 @@ func Setup(ctx context.Context, router *mux.Router, jobStore store.JobStore) *Jo
 
 	router.HandleFunc("/jobs", api.CreateJobHandler(ctx)).Methods("POST")
 	router.HandleFunc("/jobs/{id}", api.GetJobHandler(ctx)).Methods("GET")
-	router.HandleFunc("/jobs", api.GetJobsHandler(ctx)).Methods("GET")
+	router.HandleFunc("/jobs", api.GetJobsHandler)
 	return api
 }
--- a/api/api.go
+++ b/api/api.go
@@ -21,6 +21,6 @@ func Setup(ctx context.Context, router *mux.Router, jobStore store.JobStore) *Jo
 
 	router.HandleFunc("/jobs", api.CreateJobHandler(ctx)).Methods("POST")
 	router.HandleFunc("/jobs/{id}", api.GetJobHandler(ctx)).Methods("GET")
-	//router.HandleFunc("/jobs", api.GetJobsHandler(ctx)).Methods("GET")
+	router.HandleFunc("/jobs", api.GetJobsHandler(ctx)).Methods("GET")
 	return api
 }
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -88,35 +88,34 @@ func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
 	}
 }
 
-//GetJobsHandler returns a function that gets a list of existing Job resources, from the Job Store, sorted by
-//their values of last_updated time (ascending).
-func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
-	log.Event(ctx, "Creating handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.", log.INFO)
-	return func(w http.ResponseWriter, req *http.Request) {
-		ctx := req.Context()
+//GetJobsHandler gets a list of existing Job resources, from the Job Store, sorted by their values of
+//last_updated time (ascending).
+func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	log.Event(ctx, "Entering handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.", log.INFO)
 
-		//get jobs from jobStore
-		jobs, err := api.jobStore.GetJobs(ctx, sync_mux)
-		if err != nil {
-			log.Event(ctx, "getting list of jobs failed", log.Error(err), log.ERROR)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
-		}
-
-		w.Header().Set("Content-Type", "application/json")
-		jsonResponse, err := json.Marshal(jobs)
-		if err != nil {
-			log.Event(ctx, "marshalling response failed", log.Error(err), log.ERROR)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
-		}
-
-		w.WriteHeader(http.StatusOK)
-		_, err = w.Write(jsonResponse)
-		if err != nil {
-			log.Event(ctx, "writing response failed", log.Error(err), log.ERROR)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
-		}
+	//get jobs from jobStore
+	jobs, err := api.jobStore.GetJobs(ctx, sync_mux)
+	if err != nil {
+		log.Event(ctx, "getting list of jobs failed", log.Error(err), log.ERROR)
+		http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+		return
 	}
+
+	w.Header().Set("Content-Type", "application/json")
+	jsonResponse, err := json.Marshal(jobs)
+	if err != nil {
+		log.Event(ctx, "marshalling response failed", log.Error(err), log.ERROR)
+		http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	_, err = w.Write(jsonResponse)
+	if err != nil {
+		log.Event(ctx, "writing response failed", log.Error(err), log.ERROR)
+		http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+		return
+	}
+
 }
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+
 	"encoding/json"
 	"github.com/ONSdigital/log.go/log"
 	"github.com/gorilla/mux"
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -17,7 +17,7 @@ var NewID = func() string {
 
 // CreateJobHandler returns a function that generates a new Job resource containing default values in its fields.
 func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
-	log.Event(ctx, "Entering CreateJobHandler function, which generates a new Job resource.", log.INFO)
+	log.Event(ctx, "Creating handler function, which calls CreateJob and returns a new Job resource.", log.INFO)
 	return func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 
@@ -51,7 +51,7 @@ func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
 
 //GetJobHandler returns a function that gets an existing Job resource, from the Job Store, that's associated with the id passed in.
 func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
-	log.Event(ctx, "Entering GetJobHandler function, which returns an existing Job resource associated with the supplied id.", log.INFO)
+	log.Event(ctx, "Creating handler function, which calls GetJob and returns an existing Job resource associated with the supplied id.", log.INFO)
 	return func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 		vars := mux.Vars(req)
@@ -87,7 +87,7 @@ func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
 //GetJobsHandler returns a function that gets a list of existing Job resources, from the Job Store, sorted by
 //their values of last_updated time (ascending).
 func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
-	log.Event(ctx, "Entering GetJobsHandler function, which returns a list of existing Job resources held in the JobStore.", log.INFO)
+	log.Event(ctx, "Creating handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.", log.INFO)
 	return func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -83,13 +83,8 @@ func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
 	}
 }
 
-//GetJobsHandler returns a function that gets a list of existing Job resources, from the Job Store, as defined by the four query parameters passed in.
-//Those query parameters are named limit, offset, sort, and state. They are described as follows:
-//limit - The maximum number of Job resources we're returning in this Jobs resource. The default limit is 20.
-//offset - The number of Job objects into the full list that this particular response is starting at. The default number to start at is 0.
-//sort - The sort order in which to return a list of reindex job resources. This defaults to last_updated in ascending order.
-//state - Filter option to bring back specific Job objects according to their state.
-//This will be a list of comma separated values that correspond to state enum values e.g. "created,failed".
+//GetJobsHandler returns a function that gets a list of existing Job resources, from the Job Store, sorted by
+//their values of last_updated time (ascending).
 func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
 	log.Event(ctx, "Entering GetJobsHandler function, which returns a list of existing Job resources held in the JobStore.", log.INFO)
 	return func(w http.ResponseWriter, req *http.Request) {
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -96,7 +96,7 @@ func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
 		ctx := req.Context()
 
 		//get jobs from jobStore
-		jobs, err := api.jobStore.GetJobs(ctx)
+		jobs, err := api.jobStore.GetJobs(ctx, sync_mux)
 		if err != nil {
 			log.Event(ctx, "getting list of jobs failed", log.Error(err), log.ERROR)
 			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -15,6 +15,8 @@ var NewID = func() string {
 	return uuid.NewV4().String()
 }
 
+var serverErrorMessage = "internal server error"
+
 // CreateJobHandler returns a function that generates a new Job resource containing default values in its fields.
 func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
 	log.Event(ctx, "Creating handler function, which calls CreateJob and returns a new Job resource.", log.INFO)
@@ -27,7 +29,7 @@ func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
 		newJob, err := api.jobStore.CreateJob(ctx, id)
 		if err != nil {
 			log.Event(ctx, "creating and storing job failed", log.Error(err), log.ERROR)
-			http.Error(w, "Failed to create and store job", http.StatusInternalServerError)
+			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
 			return
 		}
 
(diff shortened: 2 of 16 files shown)


api/jobs.go
```diff
@@ -82,3 +82,36 @@ func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
 		}
 	}
 }
+
+//GetJobsHandler returns a function that gets a list of existing Job resources, from the Job Store, sorted by
+//their values of last_updated time (ascending).
+func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
+	log.Event(ctx, "Entering GetJobsHandler function, which returns a list of existing Job resources held in the JobStore.", log.INFO)
```
ChrysmOre, line 89: Isn't this log hit when the handler is created at startup rather than when entered?

api/jobs_test.go
```diff
@@ -155,3 +156,141 @@ func TestCreateJobHandlerWithInvalidID(t *testing.T) {
 		})
 	})
 }
+
+func TestGetJobsHandler(t *testing.T) {
+
```
ChrysmOre, line 161: I see a lot of blank lines at the beginning of functions littered throughout all the code but it's completely inconsistent.. I think it's a lot better to choose one formatting guideline and stick to it. I also think preferably the blank line should be dropped, as encouraged by gofmt/golint. After a while it _will_ stop looking awkward to your eyes, honest :)

api/jobs.go
cookel2: In terms of when the _handler_ is entered, I'm sure you're correct. But that log is the first line of the GetJobsHandler _function_ so whenever that function is called then the log will be hit (and that can only be when the function has been entered). I'm happy to move it somewhere else if you prefer. However, I would have to do the same for the other API functions as they all have logging in a similar place.

api/jobs_test.go
cookel2: It's a fair point. I have been using the gofmt tool but I guess it's leaving me to decide about the blank lines. I'm happy to remove as many as I can find.

cookel2: done

cookel2:
Hi @ChrysmOre  please would you mind reviewing again? I've removed blank lines as much as possible. I'm not sure what change I should make to the logging - please see my comment. By the way, I don't know your name? It would be good to be able to talk over Slack..

features/steps/jobs_steps.go
```diff
@@ -157,35 +178,46 @@ func (f *JobsFeature) theResponseShouldAlsoContainTheFollowingValues(table *godo
...
+	return f.ErrorFeature.StepError()
+}
+
+//callPostJobs is a utility method that can be called by a feature step in order to call the POST jobs/ endpoint
+//Calling that endpoint results in the creation of a job, in the Job Store, containing a unique id and default values.
+func (f *JobsFeature) callPostJobs() {
 	var emptyBody = godog.DocString{}
 	err := f.ApiFeature.IPostToWithBody("/jobs", &emptyBody)
 	if err != nil {
 		os.Exit(1)
```
nshumoogum, line 219: Should we not be returning err here instead of calling os.Exit?

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+}
+
+//inEachJobIWouldExpectIdLast_updatedAndLinksToHaveThisStructure is a feature step that can be defined for a specific JobsFeature.
+//It checks the response from calling GET /jobs to make sure that each job contains the expected types of values of id,
+//last_updated, and links.
+func (f *JobsFeature) inEachJobIWouldExpectIdLast_updatedAndLinksToHaveThisStructure(table *godog.Table) error {
+	assist := assistdog.NewDefault()
+	expectedResult, err := assist.ParseMap(table)
+	if err != nil {
+		panic(err)
```
nshumoogum, line 284: Should we not be returning err here instead of panic?

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+	var response models.Jobs
+
+	err = json.Unmarshal(f.responseBody, &response)
+	if err != nil {
+		return err
+	}
+
+	for j := range response.Job_List {
+		job := response.Job_List[j]
+		err2 := f.checkStructure(err, job.ID, job.LastUpdated, expectedResult, job.Links)
```
nshumoogum, line 295: I don't see it being necessary to create a new error type variable assigned to checkStructure method. We could just use existing error, err?

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+	return f.ErrorFeature.StepError()
+}
+
+//eachJobShouldAlsoContainTheFollowingValues is a feature step that can be defined for a specific JobsFeature.
+//It checks the response from calling GET /jobs to make sure that each job contains the expected values of
+//all the remaining attributes of a job.
+func (f *JobsFeature) eachJobShouldAlsoContainTheFollowingValues(table *godog.Table) error {
+	expectedResult, err := assistdog.NewDefault().ParseMap(table)
+	if err != nil {
+		panic(err)
```
nshumoogum, line 311: Again return err instead of panic(err)

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+	}
+	var response models.Jobs
+
+	_ = json.Unmarshal(f.responseBody, &response)
```
nshumoogum, line 315: We should handle the err returned from unmarshal method and return

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+
+	for j := range response.Job_List {
+		job := response.Job_List[j]
```
nshumoogum, line 318: This seems unnecessary, if we restructured the for loop like so:

```
        for _, job := range response.Job_List {
                f.checkValuesInJob(expectedResult, job)
        }
```

features/steps/jobs_steps.go
```diff
@@ -214,3 +246,103 @@ func (f *JobsFeature) iCallGETJobsidUsingTheGeneratedId() error {
...
+func (f *JobsFeature) theJobsShouldBeOrderedByLast_updatedWithTheOldestFirst() error {
+	var response models.Jobs
+	err := json.Unmarshal(f.responseBody, &response)
+	if err != nil {
+		return err
+	}
+	job_list := response.Job_List
+	timeToCheck := job_list[0].LastUpdated
+
+	j := 0
```
nshumoogum, line 337: lines 337, 338 and 345 can be reduced into 1 line to define for loop, see below (also made tweaks to other 1 liners in the for loop.

```
...
	for j := 1; j < len(job_list); j++ {
		index := strconv.Itoa(j-1)
		nextIndex := strconv.Itoa(j)
		nextTime := job_list[j].LastUpdated

		assert.True(&f.ErrorFeature, timeToCheck.Before(nextTime),
			"The value of last_updated at job_list["+index+"] should be earlier than that at job_list["+nextIndex+"]")

		timeToCheck = nextTime
	}
...
```

store/job_store.go
```diff
@@ -20,9 +23,35 @@ type DataStore struct {
...
 
+//LastUpdatedSlice is a type that implements the sort interface so that the jobs in it can be sorted using the generic Sort function.
+type LastUpdatedSlice []models.Job
+
+//Len is a function that's required by the sort interface.
+func (s LastUpdatedSlice) Len() int {
+	return len(s)
+}
+
+//Less is a function that's required by the sort interface.
```
nshumoogum, line 34: Less and Swap functions are very nice; if it wasn't for moving to using a database in the near future I would ask for some unit tests for these methods but alas I would refrain from doing so.

store/job_store.go
```diff
@@ -63,3 +91,28 @@ func (ds *DataStore) GetJob(ctx context.Context, id string) (models.Job, error)
 	log.Event(ctx, "getting job from job store", log.Data{"Job details: ": JobsMap[id]})
 	return job, nil
 }
+
+//GetJobs gets a list of Job resources from the JobsMap. It will put all the Job resources into a list, sorted by their
+//last_updated time values, and return the list.
+func (ds *DataStore) GetJobs(ctx context.Context) (models.Jobs, error) {
+	log.Event(ctx, "getting list of jobs")
```
nshumoogum, line 98: logs in this file are missing the severity level. This will not be optional in the future (via log.go library).

Please include log.INFO in the event logs (line 98, 104, 115)

nshumoogum (changes requested):

api/jobs.go
```diff
@@ -82,3 +83,36 @@ func (api *JobStoreAPI) GetJobHandler(ctx context.Context) http.HandlerFunc {
...
+func (api *JobStoreAPI) GetJobsHandler(ctx context.Context) http.HandlerFunc {
+	log.Event(ctx, "Creating handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.", log.INFO)
+	return func(w http.ResponseWriter, req *http.Request) {
+		ctx := req.Context()
+
+		//get jobs from jobStore
+		jobs, err := api.jobStore.GetJobs(ctx)
+		if err != nil {
+			log.Event(ctx, "getting list of jobs failed", log.Error(err), log.ERROR)
+			http.Error(w, "Failed to get list of jobs from job store", http.StatusInternalServerError)
```
nshumoogum, line 98: I would refrain from giving out information to the API user if they are unable to do anything about the situation.

Instead any responses containing status 500 (http.StatusInternalServerError) should just return a message of `internal server error`. This value will then be reused across the app, so you can make this a constant at the api package level.

nshumoogum (changes requested):

api/jobs.go
nshumoogum: If we give out more information this could be exploited if they were a malicious user. When our applications are responding with internal server error, we the team supporting the API will want to investigate, not the API user (nothing they can do but exploit the situation).

api/api.go
```diff
@@ -21,5 +22,6 @@ func Setup(ctx context.Context, router *mux.Router, jobStore store.JobStore) *Jo
 
 	router.HandleFunc("/jobs", api.CreateJobHandler(ctx)).Methods("POST")
 	router.HandleFunc("/jobs/{id}", api.GetJobHandler(ctx)).Methods("GET")
+	router.HandleFunc("/jobs", api.GetJobsHandler(ctx)).Methods("GET")
```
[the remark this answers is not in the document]
cookel2: I can take out the ctx, I understand what you've said, but if I do that then it gives me an error where I have used it in a log.Event just before the handler function is created. I can either remove the logging there or put it inside the handler function itself (where I can use the request context, as you've explained). What would you prefer?

api/jobs_test.go
```diff
@@ -155,3 +149,133 @@ func TestCreateJobHandlerWithInvalidID(t *testing.T) {
...
+			api.Router.ServeHTTP(resp, req)
+
+			Convey("Then a list of jobs is returned with status code 200", func() {
+				So(resp.Code, ShouldEqual, http.StatusOK)
+				payload, err := ioutil.ReadAll(resp.Body)
+				So(err, ShouldBeNil)
+				jobsReturned := models.Jobs{}
+				err = json.Unmarshal(payload, &jobsReturned)
+				expectedJob1 := models.NewJob(testJobID1)
```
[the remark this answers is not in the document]
cookel2: I see what you mean. I've created an ExpectedJob method to use instead of the NewJob method (for the tests) now.

ChrysmOre (changes requested):

features/steps/jobs_steps.go
cookel2: done

features/steps/jobs_steps.go
cookel2: yes I like that idea - done

features/steps/jobs_steps.go
cookel2: yes that looks better - I should have thought about it more because it was a little messy before - done

store/job_store.go
cookel2: thanks : )

api/jobs.go
cookel2: ok, that makes sense - done

store/job_store.go
cookel2: I've added it now

features/steps/jobs_steps.go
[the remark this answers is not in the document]
cookel2: resolved earlier

models/jobs.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+//Jobs represents an array of Job resources and json representation for API
+type Jobs struct {
+	Job_List []Job `json:"jobs"`
```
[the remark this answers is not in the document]
cookel2: I'll need to check the DP coding standards to make sure that they follow Go's conventions - I wouldn't like to have to change them back again later on!

store/job_store.go
```diff
@@ -20,9 +23,35 @@ type DataStore struct {
...
+func (s LastUpdatedSlice) Less(i, j int) bool {
+	return s[i].LastUpdated.Before(s[j].LastUpdated)
+}
+
+//Swap is a function that's required by the sort interface.
+func (s LastUpdatedSlice) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+//DeleteAllJobs empties the JobStore by deleting everything from the JobsMap.
```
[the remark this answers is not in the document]
cookel2: I'll read up - thanks

features/steps/jobs_steps.go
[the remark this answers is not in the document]
cookel2: The panicking is a hang over from Armakuni's code - I'll change it.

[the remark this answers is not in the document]
cookel2: done

models/jobs.go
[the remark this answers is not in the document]
ChrysmOre: Yeah, sounds like a good idea.

api/api.go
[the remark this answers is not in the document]
ChrysmOre: If the logging isn't there for a specific purpose it begs the question of if it's really necessary in the first place? Another solution is just to pass context.TODO() to the logger (it doesn't seem like it particularly needs any details from the context).
If you do remove the logging and the context then there's no reason to have a function that returns a function, you can just use the Handlerfunc directly.

store/job_store.go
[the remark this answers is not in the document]
cookel2: I've read up and it seems like a good idea to add the use of a sync.Mutex to all the functions that access the jobs map. So I've made those changes now.

api/api.go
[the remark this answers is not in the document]
cookel2: I've now removed the context and changed it to use the HandlerFunc directly, as suggested by @ChrysmOre. I've changed the logging so that it shows when the handler function is being entered (my previous tech lead taught me that this is good practice).

api/jobs_test.go
```diff
@@ -155,3 +149,133 @@ func TestCreateJobHandlerWithInvalidID(t *testing.T) {
...
+					So(len(returnedJobList), ShouldEqual, 0)
+				})
+			})
+		})
+	})
+}
+
+func TestGetJobsHandlerWithInternalServerError(t *testing.T) {
+	t.Parallel()
+	Convey("Given a Search Reindex Job API that generates an internal server error", t, func() {
```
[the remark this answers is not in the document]
cookel2: I've removed the result from the wording now.

features/steps/jobs_steps.go
```diff
@@ -129,25 +136,39 @@ func (f *JobsFeature) iWouldExpectIdLast_updatedAndLinksToHaveThisStructure(tabl
...
+	if err2 != nil {
+		return err2
+	}
+
+	return f.ErrorFeature.StepError()
+}
+
+//checkStructure is a utility method that can be called by a feature step to assert that a job contains the expected structure in its values of
+//id, last_updated, and links. It confirms that last_updated is a current or past time, and that the tasks and self links have the correct paths.
+func (f *JobsFeature) checkStructure(err error, id string, lastUpdated time.Time, expectedResult map[string]string, links *models.JobLinks) error {
```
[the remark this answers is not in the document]
cookel2: I've removed the err from the input parameters.

features/steps/jobs_steps.go
```diff
@@ -129,25 +136,39 @@ func (f *JobsFeature) iWouldExpectIdLast_updatedAndLinksToHaveThisStructure(tabl
 		return err
 	}
 
-	_, err = uuid.FromString(response.ID)
+	id := response.ID
+	lastUpdated := response.LastUpdated
+	links := response.Links
+
+	err2 := f.checkStructure(err, id, lastUpdated, expectedResult, links)
```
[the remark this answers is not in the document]
cookel2: Yes it can - I've changed it now.

features/steps/jobs_steps.go
```diff
@@ -129,25 +136,39 @@ func (f *JobsFeature) iWouldExpectIdLast_updatedAndLinksToHaveThisStructure(tabl
...
 	}
 
-	linksTasks := strings.Replace(expectedResult["links: tasks"], "{id}", response.ID, 1)
+	linksTasks := strings.Replace(expectedResult["links: tasks"], "{id}", id, 1)
 
-	assert.Equal(&f.ErrorFeature, linksTasks, response.Links.Tasks)
+	assert.Equal(&f.ErrorFeature, linksTasks, links.Tasks)
 
-	linksSelf := strings.Replace(expectedResult["links: self"], "{id}", response.ID, 1)
+	linksSelf := strings.Replace(expectedResult["links: self"], "{id}", id, 1)
```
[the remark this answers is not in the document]
cookel2: now renamed

features/steps/jobs_steps.go
cookel2: yes I agree - changed now

features/steps/jobs_steps.go
cookel2: yes - done

nshumoogum (dismissed):

ChrysmOre (changes requested):

store/job_store.go
```diff
@@ -34,19 +66,20 @@ func (ds *DataStore) CreateJob(ctx context.Context, id string) (models.Job, erro
 	newJob := models.NewJob(id)
 
 	//Check that the JobsMap does not already contain the id as a key
+	mux.Lock()
```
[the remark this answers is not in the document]
cookel2: I hadn't thought of doing it that way but it makes sense - done

store/job_store.go
```diff
@@ -55,11 +88,42 @@ func (ds *DataStore) GetJob(ctx context.Context, id string) (models.Job, error)
 	}
 
 	//Check that the JobsMap contains the id as a key
+	mux.Lock()
 	if _, idPresent := JobsMap[id]; idPresent == false {
```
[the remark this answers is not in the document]
cookel2: done

store/job_store.go
```diff
@@ -20,9 +24,37 @@ type DataStore struct {
...
+	mux.Lock()
+	for k := range JobsMap {
+		delete(JobsMap, k)
+	}
+	mux.Unlock()
+	return nil
+}
 
+// CreateJob creates a new Job resource and stores it in the JobsMap.
+func (ds *DataStore) CreateJob(ctx context.Context, id string, mux *sync.Mutex) (models.Job, error) {
```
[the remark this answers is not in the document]
cookel2: I agree it would be worth it except that my next ticket will require me to remove all the Mutex code as I will be replacing the map with calls to Mongo dB. So I would prefer to leave things as they are. Is that ok?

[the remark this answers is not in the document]
ChrysmOre: Ah ok sure, no worries

ChrysmOre (approved):

PR #11, opened 2021-05-13
Add details of health endpoint to the swagger

cookel2:
# What has changed

The GET /health endpoint has been added to the swagger.

The health endpoint returns 200 status OK response with no checks on dependent services.

# How to review

Check that the details in the swagger spec look correct. 

Use the Swagger Editor to check that there are no errors: https://editor.swagger.io/

# Who can review

Anyone but me

### DIFF
--- a/swagger.yaml
+++ b/swagger.yaml
@@ -18,6 +18,23 @@ tags:
 
 paths:
 
+  /health:
+    get:
+      tags:
+        - Public
+      summary: "Get health of Search Reindex API"
+      description:
+        "Returns health information of Search Reindex API and list of the health of its dependent services"
+      produces:
+        - "application/json"
+      responses:
+        200:
+          description: "Health information and list of dependent services health. NB. It currently has no dependent services so the list is empty."
+          schema:
+            $ref: "#/definitions/Health"
+        500:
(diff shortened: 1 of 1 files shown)


CarlHembrough (approved):

CarlHembrough (commented):
Consider also adding detail for the 429 HTTP code, when the healthcheck returns a warning. This can happen when the service first starts and has not yet run the checks for the first time. Though this may not occur while you have no checkers registered.

https://github.com/ONSdigital/dp-healthcheck/blob/bb901bf57f3bd3c29c53a064844079fec75d6c1f/healthcheck/handler.go#L33

cookel2:
> Consider also adding detail for the 429 HTTP code, when the healthcheck returns a warning. This can happen when the service first starts and has not yet run the checks for the first time. Though this may not occur while you have no checkers registered.
> 
> https://github.com/ONSdigital/dp-healthcheck/blob/bb901bf57f3bd3c29c53a064844079fec75d6c1f/healthcheck/handler.go#L33

Thanks for this idea Carl. I'll look at doing that when I add a check for mongo dB (in a future ticket).

PR #13, opened 2021-06-09
Feature/use persistent datastore

cookel2:
# What has changed

The dp-search-reindex-api no longer uses a Map, for storing details of search reindex jobs, but has now been changed to use MongoDB (a persistent datastore) instead. 

The jobs are now stored, in MongoDB, as documents within a collection named 'jobs', which sits within a database named 'search'.

The unit tests and component tests have also been updated to use mongoDB as necessary.

# How to review

The endpoints can be tested locally by first running MongoDB locally as follows:
- pull down the latest version of dp-compose and go into it's root directory. Then run these commands:
- docker-compose up -d
- docker start dp-compose_mongodb_1
- mongo

There should now be a local connection to mongodb://127.0.0.1:27017

The functionality can now be tested by moving into the root of dp-search-reindex-api, pulling down this branch, then running this command: make debug
Postman can be used to test that the following endpoints all still work as defined in the swagger:
- POST: http://localhost:25700/jobs (should post a default job into mongoDB and return a JSON representation of it)
- GET: http://localhost:25700/jobs/id NB. Use the id returned by the POST call above e.g. http://localhost:25700/jobs/bc7b87de-abf5-45c5-8e3c-e2a575cab28a (should get the relevant job from mongoDB)
- GET: http://localhost:25700/jobs (should get all the jobs from mongoDB)
- GET: http://localhost:25700/health (should show the health check details)

# Who can review

Anyone but me


mongo/mgo_job_store.go
```diff
@@ -0,0 +1,198 @@
...
+	dpMongoLock "github.com/ONSdigital/dp-mongodb/dplock"
+	dpMongoHealth "github.com/ONSdigital/dp-mongodb/health"
+	"github.com/ONSdigital/dp-search-reindex-api/models"
+	"github.com/ONSdigital/log.go/log"
+	"github.com/globalsign/mgo"
+	"github.com/globalsign/mgo/bson"
+)
+
+// MgoJobStore defines the required methods from MongoDB
+type MgoJobStore interface {
```
jessjenkins, line 16: Similarly to the filename, this is a `JobStore` in the `mongo` so the Mgo prefix is not needed. Also there's no need to specify the library we are using as part of the name. If we choose to use something other than `mgo` in future, we wouldn't want to have to change this package's interface or have it being misleading.
```suggestion
type JobStorer interface {
```

Additionally, the interface belongs in the package that is using it in go. This is different from Java where the interface is defined by the package that owns the implementation in many cases. This is due to the way interfaces are defined in go. So for instance…
In Java: A class might say 'I am xyz and I implement interface abc'
In Go: A package might say 'I need something that satisfies the interface I am defining called abc'. The other type doesn't need to explicitly define that it satisfies that interface it just has to have all the required functions.

For this reason I'd move the interface to where it's used. Also it's OK for different packages to define different interfaces for only the things they need.

mongo/mgo_job_store.go
```diff
@@ -0,0 +1,198 @@
...
+//Less is a function that's required by the sort interface.
+func (s LastUpdatedSlice) Less(i, j int) bool {
+	return s[i].LastUpdated.Before(s[j].LastUpdated)
+}
+
+//Swap is a function that's required by the sort interface.
+func (s LastUpdatedSlice) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
```
jessjenkins, line 42: These aren't needed any more since Mongo is sorting the documents now.

mongo/mgo_job_store.go
```diff
@@ -0,0 +1,198 @@
...
+// jobs collection name
+const jobsCol = "jobs"
+
+// locked jobs collection name
+const jobsLockCol = "jobs_locks"
+
+//MgoDataStore is a type that contains an implementation of the MgoJobStore interface, which can be used for creating and getting Job resources.
+//It also represents a simplistic MongoDB configuration, with session, health and lock clients
+type MgoDataStore struct {
```
jessjenkins, line 51: Mgo not needed, also I'd be more specific about the kind of data being stored.
```suggestion
type JobStore struct {
```

mongo/mongotest/mgo_job_store_test.go
```diff
@@ -1,8 +1,8 @@
-package store
+package mongotest
```
jessjenkins, line 1: This test file should be renamed and moved to the parent directory so the test is in the same dir as the thing it's testing.
The filename should then be consistent but with `_test.go`. I'd suggest the package code being called `job_store.go` with this file being called `job_store_test.go`. The `package` in this file should then be `mongo_test` which is a special package name that allows tests to be in the same dir as the code.
```suggestion
package mongo_test
```

For an example of this from another service, look at [the api package in the dp-image-api service](https://github.com/ONSdigital/dp-image-api/blob/develop/api/image_test.go#L1).

api/jobs.go
```diff
@@ -119,3 +125,10 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
 	}
 
 }
+
+//unlockJob unlocks the provided job lockID and logs any error with WARN state
+func (api *JobStoreAPI) unlockJob(ctx context.Context, lockID string) {
+	if err := api.mongoDB.UnlockJob(lockID); err != nil {
+		log.Event(ctx, "error unlocking mongoDB lock for an image resource", log.WARN, log.Data{"lockID": lockID})
```
jessjenkins, line 132: This isn't an image api.

api/api.go
```diff
@@ -2,26 +2,32 @@ package api
...
+	"github.com/ONSdigital/log.go/log"
 	"github.com/gorilla/mux"
 )
 
 //JobStoreAPI provides a struct to wrap the api around
 type JobStoreAPI struct {
-	Router   *mux.Router
-	jobStore store.JobStore
+	Router  *mux.Router
+	mongoDB mongo.MgoJobStore
```
jessjenkins, line 13: I'd have kept this is a JobStore, the api package doesn't really care that it's storing the jobs in mongo, it only cares that the jobs are stored and can be retrieved.

jessjenkins (commented):
Made some comments, I don't want to overwhelm with too many. Once addressed we can go through it again.

service/interfaces.go
```diff
@@ -2,20 +2,25 @@ package service
...
 	"github.com/ONSdigital/dp-healthcheck/healthcheck"
 	"github.com/ONSdigital/dp-search-reindex-api/config"
+	"github.com/ONSdigital/dp-search-reindex-api/mongo"
+	"net/http"
 )
 
 //go:generate moq -out mock/initialiser.go -pkg mock . Initialiser
 //go:generate moq -out mock/server.go -pkg mock . HTTPServer
 //go:generate moq -out mock/healthCheck.go -pkg mock . HealthChecker
+//go:generate moq -out mock/mgo_job_store.go -pkg mock . MongoServer
```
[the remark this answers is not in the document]
cookel2: thanks for pointing this out - I've removed it now

service/service.go
```diff
@@ -2,52 +2,56 @@ package service
...
 	"github.com/ONSdigital/dp-search-reindex-api/config"
+	"github.com/ONSdigital/dp-search-reindex-api/mongo"
 	"github.com/ONSdigital/log.go/log"
 	"github.com/gorilla/mux"
+	"github.com/justinas/alice"
 	"github.com/pkg/errors"
 )
 
-// Service contains all the configs, server and clients to run the dp-search-reindex-api API
+// Service contains all the configs, server and clients to run the Image API
```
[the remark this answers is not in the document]
cookel2: Oops! I've been copying your code of course! - corrected now

service/service_test.go
```diff
@@ -3,18 +3,18 @@ package service_test
 import (
 	"context"
 	"fmt"
+	"github.com/ONSdigital/dp-search-reindex-api/mongo"
```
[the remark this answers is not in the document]
cookel2: done

mongo/mgo_job_store.go
```diff
@@ -0,0 +1,198 @@
+package mongo
```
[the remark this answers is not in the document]
cookel2: yes I agree - I had named it with the mgo prefix to make things clearer to me while the old job_store class, which used a Map rather than Mongo, was still there (but I've deleted that now). - prefix now removed

mongo/mgo_job_store.go
cookel2: I've removed them now.

api/api.go
cookel2: done

mongo/mgo_job_store.go
cookel2: done

api/jobs.go
cookel2: corrected now

mongo/mongotest/mgo_job_store_test.go
cookel2: that's interesting to know - I've changed it now - thanks

api/api.go
```diff
@@ -2,26 +2,31 @@ package api
...
+	"github.com/ONSdigital/log.go/log"
 	"github.com/gorilla/mux"
 )
 
 //JobStoreAPI provides a struct to wrap the api around
 type JobStoreAPI struct {
-	Router   *mux.Router
-	jobStore store.JobStore
+	Router  *mux.Router
+	mongoDB JobStorer
```
jessjenkins, line 12: I'd still call this field `jobStore` as the `api` package doesn't really care where the job is stored to, it just wants to store the job somewhere.

api/api.go
```diff
@@ -2,26 +2,31 @@ package api
...
 }
 
 //Setup function sets up the api and returns an api
-func Setup(ctx context.Context, router *mux.Router, jobStore store.JobStore) *JobStoreAPI {
+func Setup(ctx context.Context, router *mux.Router, mongoDB JobStorer) *JobStoreAPI {
 	api := &JobStoreAPI{
-		Router:   router,
-		jobStore: jobStore,
+		Router:  router,
+		mongoDB: mongoDB,
```
jessjenkins, line 19: As above, I'd remove references to mongo from the `api` package and stick to the more generic `jobStore`

api/api.go
```diff
@@ -2,26 +2,31 @@ package api
...
 
 	router.HandleFunc("/jobs", api.CreateJobHandler(ctx)).Methods("POST")
 	router.HandleFunc("/jobs/{id}", api.GetJobHandler(ctx)).Methods("GET")
 	router.HandleFunc("/jobs", api.GetJobsHandler)
 	return api
 }
+
+//Close is called during graceful shutdown to give the API an opportunity to perform any required disposal task
+func (*JobStoreAPI) Close(ctx context.Context) error {
+	log.Event(ctx, "graceful shutdown of JobStore Api complete", log.INFO)
```
jessjenkins, line 30: This is a search-reindex api rather than a "jobstore api".  The job store is just a dependency of the api.
```suggestion
	log.Event(ctx, "graceful shutdown of api complete", log.INFO)
```

api/api_test.go
```diff
@@ -4,18 +4,16 @@ import (
...
 )
 
 func TestSetup(t *testing.T) {
 	Convey("Given an API instance", t, func() {
-		r := mux.NewRouter()
-		ctx := context.Background()
-		js := &store.DataStore{}
-		api := Setup(ctx, r, js)
+
+		api := Setup(context.Background(), mux.NewRouter(), &mongo.JobStore{})
```
jessjenkins, line 16: I'd use a mock of the jobstore (api/mock.JobStorerMock) rather than the actual mongo package (sticking to the ideal of api not knowing anything about mongo specifically)

api/api_test.go
```diff
@@ -25,6 +23,18 @@ func TestSetup(t *testing.T) {
 	})
 }
 
+func TestClose(t *testing.T) {
+	Convey("Given an API instance", t, func() {
+		ctx := context.Background()
+		api := Setup(context.Background(), mux.NewRouter(), &mongo.JobStore{})
```
jessjenkins, line 29: Again, use a jobstorer mock instead of the mongo package.

api/jobs.go
```diff
@@ -27,8 +24,8 @@ func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
 
 		id := NewID()
 
-		//Create job in job store
-		newJob, err := api.jobStore.CreateJob(ctx, id, sync_mux)
+		//Create job in jobs collection in mongoDB
+		newJob, err := api.mongoDB.CreateJob(ctx, id)
```
jessjenkins, line 28: Again remove references to mongoDB from the API package.
```suggestion
		//Create job in job store
		newJob, err := api.jobStore.CreateJob(ctx, id)
```

api/jobs.go
```diff
@@ -94,8 +100,8 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
 	ctx := req.Context()
 	log.Event(ctx, "Entering handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.", log.INFO)
 
-	//get jobs from jobStore
-	jobs, err := api.jobStore.GetJobs(ctx, sync_mux)
+	//get jobs from jobs collection in mongoDB
```
jessjenkins, line 103: Remove mongo DB from api package.

api/jobs_test.go
```diff
@@ -32,11 +32,15 @@ var ctx = context.Background()
...
+	api.NewID = func() string { return testJobID1 }
+
+	jobsCollectionMock := &apiMock.JobStorerMock{
+		CreateJobFunc: func(ctx context.Context, id string) (models.Job, error) { return models.NewJob(id), nil },
+	}
 
 	Convey("Given a Search Reindex Job API that can create valid search reindex jobs and store their details in a map", t, func() {
-		api := Setup(ctx, mux.NewRouter(), &store.DataStore{})
-		createJobHandler := api.CreateJobHandler(ctx)
+		jobStoreApi := api.Setup(ctx, mux.NewRouter(), jobsCollectionMock)
```
jessjenkins, line 42: Not a jobStore Api so this name is misleading. I'd rename, perhaps to `reindexApi` or even `apiInstance` to differentiate it from the api package???

api/jobs_test.go
```diff
@@ -73,8 +77,8 @@ func TestCreateJobHandlerWithValidID(t *testing.T) {
 func TestGetJobHandler(t *testing.T) {
 	t.Parallel()
 	Convey("Given a Search Reindex Job API that returns specific jobs using their id as a key", t, func() {
-		jobStoreMock := &mock.JobStoreMock{
-			GetJobFunc: func(ctx context.Context, id string, mux *sync.Mutex) (models.Job, error) {
+		jobsCollectionMock := &apiMock.JobStorerMock{
```
jessjenkins, line 80: I'd probably leave this as `jobStoreMock`

api/jobs_test.go
jessjenkins: NB. I haven't flagged all instances of where this'll change but there are more examples below…

api/jobs_test.go
jessjenkins: Not flagged all instances, more below…

features/steps/jobs_steps.go
```diff
@@ -10,36 +10,42 @@ import (
...
-	"github.com/ONSdigital/dp-search-reindex-api/service/mock"
+	serviceMock "github.com/ONSdigital/dp-search-reindex-api/service/mock"
+	"github.com/benweissmann/memongo"
 	"github.com/cucumber/godog"
 	"github.com/pkg/errors"
 	"github.com/rdumont/assistdog"
 	uuid "github.com/satori/go.uuid"
 	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"net/http"
```
jessjenkins, line 27: I'd move these two standard library imports up to the first block of imports (after "time").

models/job.go
```diff
@@ -6,7 +6,8 @@ import (
 
 //Job represents a job metadata model and json representation for API
 type Job struct {
-	ID                           string    `json:"id"`
+	ID                           string    `bson:"_id" json:"id"`
+	CollectionID                 string    `bson:"collection_id,omitempty" json:"collection_id,omitempty"`
```
jessjenkins, line 10: I'd add bson values for the other fields in the struct too. Not strictly necessary if they match the field names but it helps with consistence should any refactoring take place.

mongo/job_store.go
```diff
@@ -0,0 +1,170 @@
...
+	dpMongoLock "github.com/ONSdigital/dp-mongodb/dplock"
+	dpMongoHealth "github.com/ONSdigital/dp-mongodb/health"
+	"github.com/ONSdigital/dp-search-reindex-api/models"
+	"github.com/ONSdigital/log.go/log"
+	"github.com/globalsign/mgo"
+	"github.com/globalsign/mgo/bson"
+)
+
+// jobs collection name
+const jobsCol = "jobs"
```
jessjenkins, line 17: The collection name is defined in the config. Maybe we should use that instead of a hard-coded constant?

mongo/job_store.go
```diff
@@ -0,0 +1,170 @@
...
+	return m.lockClient.Unlock(lockID)
+}
+
+// Close closes the mongo session and returns any error
+func (m *JobStore) Close(ctx context.Context) error {
+	m.lockClient.Close(ctx)
+	return dpMongodb.Close(ctx, m.Session)
+}
+
+func (m *JobStore) GetJobs(ctx context.Context) (models.Jobs, error) {
```
jessjenkins, line 116: All exported functions, structs etc. need GoDoc comments.

mongo/job_store.go
```diff
@@ -0,0 +1,170 @@
...
+		Client:      *m.client,
+		Healthcheck: m.client.Healthcheck,
+	}
+
+	// Create MongoDB lock client, which also starts the purger loop
+	m.lockClient = dpMongoLock.New(ctx, m.Session, m.Database, jobsCol)
+	return nil
+}
+
+//AcquireJobLock tries to lock the provided jobID.
```
jessjenkins, line 98: Minor formatting suggestion, some comments have a space between the `//` and the text, I think it looks cleaner.

mongo/job_store.go
```diff
@@ -0,0 +1,170 @@
...
+		return results, err
+	}
+
+	results.JobList = jobs
+	log.Event(ctx, "list of jobs - sorted by last_updated", log.Data{"Sorted jobs: ": results.JobList}, log.INFO)
+
+	return results, nil
+}
+
+func (m *JobStore) GetJob(ctx context.Context, id string) (models.Job, error) {
```
jessjenkins, line 150: GoDoc needed

api/api.go
cookel2: makes sense - done

api/api.go
cookel2: amended now

api/api.go
cookel2: makes sense - done

api/api_test.go
cookel2: thanks for pointing this out - done

api/jobs.go
cookel2: now changed

api/api_test.go
cookel2: good idea - done

api/jobs.go
cookel2: done

api/jobs_test.go
cookel2: fair enough - I've named it apiInstance now

api/jobs_test.go
cookel2: name reverted back to jobStoreMock now

features/steps/jobs_steps.go
cookel2: thanks for pointing this out

mongo/job_store.go
cookel2: I had a little look at doing it like that but this is similar to what you have done in the dp-image-api, which seems to save some lines of code this way.

models/job.go
cookel2: I've added the bson values and also removed the CollectionID attribute (I must have added this early on when I was unsure how to use mongo)!

mongo/job_store.go
cookel2: more GoDoc added now

mongo/job_store.go
cookel2: done

mongo/job_store.go
cookel2: done

mongo/job_store.go
cookel2: I've changed the code as discussed - jobs_locks is now configurable too.

jessjenkins (approved):

PR #19, opened 2021-07-08
Feature/add endpoint to put count tasks

cookel2:
# What has changed

An endpoint has been added that allows a PUT request to be made, for a specific search reindex job, which passes in the job id and a count as path parameters. The count parameter is used to update the number_of_tasks field for the relevant Job resource. A status of 200 is returned if successful.

Unit tests have been added for the api and the handler.

A component test has also been added, which tests that the Job resource number_of_tasks field will be correctly updated when the PUT endpoint is called.

# How to review

Run mongo DB locally. Run the code locally using this command: make debug

The endpoint can then be tested locally by first using this POST URL to create a Job resource in the Job Store: http://localhost:25700/jobs
The id can then be copied from the returned Job resource and used in the PUT URL http://localhost:25700/jobs/{id}/number_of_tasks/{count}
The number_of_tasks value can be checked by looking at the Job resource returned by the GET URL for the same id http://localhost:25700/jobs/{id}

# Who can review

Anyone but me


api/jobs.go
```diff
@@ -119,12 +120,51 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+func (api *JobStoreAPI) PutNumTasksHandler(ctx context.Context) http.HandlerFunc {
+	log.Event(ctx, "Creating handler function, which calls PutNumberOfTasks to update the number_of_tasks in the job with the supplied id.", log.INFO)
+	return func(w http.ResponseWriter, req *http.Request) {
+		ctx := req.Context()
+		vars := mux.Vars(req)
+		id := vars["id"]
+		count := vars["count"]
+
+		var numTasks int
+		numTasks, err := strconv.Atoi(count)
```
daiLlew, line 143: You don't need to declare this and then assign it. You can just do:
```
numTasks, err := strconv.Atoi(count)
```

api/jobs.go
```diff
@@ -119,12 +120,51 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+		logData := log.Data{"job_id": id, "num_tasks": numTasks}
```
daiLlew, line 144: Minor point - would it be more beneficial to do 
```
log.Data{"job_id": id, "num_tasks": count}
```
If count is not a valid int or cannot be parsed to an int then `numTasks` will be the zero value for it so it will hide what the offending value was. If you add count instead you will be able to see what the value was which might be useful if this request fails with an error. As I said really minor point but could be really helpful in the right circumstances.

api/jobs.go
```diff
@@ -119,12 +120,51 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+			log.Event(ctx, "acquiring lock for job ID failed", log.Error(err), logData, log.ERROR)
+			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+			return
+		}
+		defer api.unlockJob(ctx, lockID)
+
+		err = api.jobStore.PutNumberOfTasks(req.Context(), id, numTasks)
+		if err != nil {
+			log.Event(ctx, "putting number of tasks failed", log.Error(err), logData, log.ERROR)
+			http.Error(w, "Failed to find job in job store", http.StatusNotFound)
```
daiLlew, line 163: Is `"Failed to find job in job store"` accurate for this error? Also should is not found status the most appropriate state code for this scenario?

api/jobs_test.go
```diff
@@ -316,3 +341,80 @@ func ExpectedJob(id string,
...
+			resp := httptest.NewRecorder()
+
+			apiInstance.Router.ServeHTTP(resp, req)
+
+			Convey("Then an error with status code 500 is returned", func() {
+				So(resp.Code, ShouldEqual, http.StatusInternalServerError)
+				errMsg := strings.TrimSpace(resp.Body.String())
+				So(errMsg, ShouldEqual, expectedServerErrorMsg)
+			})
+		})
```
daiLlew, line 418: Question: what happens if count is negative? Granted its unlikely but that value is provided to the service so its wise to validate its OK rather than assume its OK. 

If setting a negative value would be considered incorrect or invalid can we add another test case to prove this is handled correctly?

mongo/job_store.go
```diff
@@ -173,3 +174,29 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (models.Job, error) {
...
+
+	return err
+}
+
+// UpdateJob updates a particular job with the values passed in through the 'updates' input parameter
+func (m *JobStore) UpdateJob(updates bson.M, s *mgo.Session, id string) error {
+	update := bson.M{"$set": updates}
+	if err := s.DB(m.Database).C(m.Collection).UpdateId(id, update); err != nil {
+		if err == mgo.ErrNotFound {
+			return errors.New("the job id could not be found in the jobs collection")
```
daiLlew, line 197: Question:
Is job not found a state that external packages would need to know about? If so consider making a public package error var for this - It provides the caller with the ability to check for this specific error and act on it if they need to.

api/jobs.go
cookel2: good point - changed now - thanks

api/jobs.go
cookel2: I agree this could be useful, I had forgotten that I could put an int in the log data, thanks for the idea.

api/jobs.go
cookel2: Good spot. I think this was laziness on my part. I've updated it now so that the "Failed to find job in job store" message will only occur if that is definitely what has gone wrong. Otherwise it will now give an internal server error if something unexpected occurs. I've updated the unit tests for this too.

api/jobs_test.go
cookel2: good idea - done

mongo/job_store.go
cookel2: I'm not sure whether "job not found" is a state that external packages might need to be aware of or not to be honest. My understanding is that this PUT endpoint will only ever get called by a service, rather than by a human, so does that make a difference?

daiLlew: Update job can return many different types of error the majority of which would mean returning a 500 status.  There are some cases where the caller will want be able to handle certain errors in a certain way for example:
- Duplicate record
- User not authorised
- Record not found
- etc.

By allowing the caller to identify which error they have it allows them to handle it in the most appropriate way
i.e. return a 404 for error == job not found.

We can do the same by returning a error like this:
```
return errors.New("job not found")
```
**But** if you were to do this then the caller would have to look inside this package at the private internal workings to know what error text is used for each case. This is bad because by writing code with knowledge of the internal workings of another package you are now tightly coupled to it. If that error text were to change in future your code would fail and you would have no warning of it. You might not own that code either so it could be updated or changed and you would have no control over it.

By making public error vars we allow the caller to make the same checks on an error returned without having to look at the inner details of the package. Also by making them public they cannot be changed without breaking backwards compatibility - essentially we are making a promise on how this package will work and we have to adhere to it.

Its fairly common pattern in Go the `mgo` package you're using here does it and the [io package](https://golang.org/pkg/io/#PipeReader.Read) below uses public vars to indicate specific error scenarios

cookel2: Thanks David, I think I get it now, I've changed it to do this.

api/jobs.go
```diff
@@ -119,12 +123,58 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+			log.Event(ctx, "acquiring lock for job ID failed", log.Error(err), logData, log.ERROR)
+			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+			return
+		}
+		defer api.unlockJob(ctx, lockID)
+
+		err = api.jobStore.PutNumberOfTasks(req.Context(), id, numTasks)
+		if err != nil {
+			log.Event(ctx, "putting number of tasks failed", log.Error(err), logData, log.ERROR)
+			if err.Error() == "the job id could not be found in the jobs collection" {
```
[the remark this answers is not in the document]
cookel2: good idea - done

api/jobs.go
```diff
@@ -119,12 +123,58 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+		id := vars["id"]
+		count := vars["count"]
+		logData := log.Data{"id": id, "count": count}
+
+		numTasks, err := strconv.Atoi(count)
+		floatNumTasks := float64(numTasks)
+		isNegative := math.Signbit(floatNumTasks)
+		if isNegative {
+			err = errors.New("the count is negative")
+		}
```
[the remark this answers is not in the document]
cookel2: I know what you mean but I was thinking that if an error did occur after the line that says: numTasks, err := strconv.Atoi(count) Then, whatever else happened in the following 2 lines, the following check (if err != nil) already there would find numTasks to be either negative or invalid in some other way. Either way the correct error message would go out. Having said that, I've updated the code as you've suggested anyway.

api/jobs.go
```diff
@@ -15,6 +19,7 @@ var NewID = func() string {
 }
 
 var serverErrorMessage = "internal server error"
+var ErrJobNotFound = "the job id could not be found in the jobs collection"
```
daiLlew, line 22: This should belong to the job store package - sorry I should have been clearer in my previous comment.

api/jobs.go
```diff
@@ -119,12 +124,64 @@ func (api *JobStoreAPI) GetJobsHandler(w http.ResponseWriter, req *http.Request)
...
+		ctx := req.Context()
+		vars := mux.Vars(req)
+		id := vars["id"]
+		count := vars["count"]
+		logData := log.Data{"id": id, "count": count}
+
+		numTasks, err := strconv.Atoi(count)
+		if err != nil {
+			log.Event(ctx, "invalid path parameter - count should be a positive integer", log.Error(err), logData, log.ERROR)
+			http.Error(w, "invalid path parameter - count should be a positive integer", http.StatusBadRequest)
```
daiLlew, line 150: "failed to convert count to integer"

mongo/job_store.go
```diff
@@ -173,3 +175,29 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (models.Job, error) {
...
+
+	return err
+}
+
+// UpdateJob updates a particular job with the values passed in through the 'updates' input parameter
+func (m *JobStore) UpdateJob(updates bson.M, s *mgo.Session, id string) error {
+	update := bson.M{"$set": updates}
+	if err := s.DB(m.Database).C(m.Collection).UpdateId(id, update); err != nil {
+		if err == mgo.ErrNotFound {
+			return apierrors.ErrJobNotFound
```
daiLlew, line 198: This error should belong to this package - the store package shouldn't depend on the API as this will create a cyclic dependency.

api/jobs.go
cookel2: I've changed the message as suggested. I had left it as it was because, although the count would not be found negative at this point, it is always the case that the count needs to be a positive integer. However, having said that, I do agree that it looked a little strange and I had previously thought about just saying "count should be an integer" instead.

mongo/job_store.go
cookel2: ok - changed now, thanks

api/jobs.go
cookel2: ok, done

daiLlew (approved):

PR #22, opened 2021-07-21
Feature/check no inflight jobs

cookel2:
# What has changed

As an API user
I do not want to trigger multiple reindexes in a short space of time
so that the backend processing of a reindex job has time to successfully finish before a new reindex job can be triggered

These are the changes:

The endpoint that allows a POST request to be made, which creates a search reindex job document in mongoDB, has been amended so that the new job will only get created if there are no jobs currently in progress.

If a job is currently in progress then it will contain the following value of state:

"state": "in-progress"

So the endpoint now creates an iteration of all the jobs that have a state set to "in-progress" and orders it so that the one with the most recent reindex_started time is first.

It uses a new environment variable named MAX_REINDEX_JOB_RUNTIME, which gives a duration in seconds, and subtracts this from the current time to calculate a "check from" time.

It then loops through the iteration and checks each reindex_started time to see if it falls between the "check from" time and now. If it does then a 409 conflict error occurs with the following message:

"existing reindex job in progress"

A unit test has been added for the POST request handler.

A few other changes have been made to refactor existing code and increase the coverage of both unit tests and component tests.

# How to review

The endpoint can be tested locally using this POST URL: http://localhost:25700/jobs

# Who can review

Anyone but me

### DIFF
--- a/mongo/job_store.go
+++ b/mongo/job_store.go
@@ -9,6 +9,7 @@ import (
 	dpMongodb "github.com/ONSdigital/dp-mongodb"
 	dpMongoLock "github.com/ONSdigital/dp-mongodb/dplock"
 	dpMongoHealth "github.com/ONSdigital/dp-mongodb/health"
+	"github.com/ONSdigital/dp-search-reindex-api/config"
 	"github.com/ONSdigital/dp-search-reindex-api/models"
 	"github.com/ONSdigital/log.go/log"
 	"github.com/globalsign/mgo"
@@ -27,6 +28,7 @@ type JobStore struct {
 	client          *dpMongoHealth.Client
 	healthClient    *dpMongoHealth.CheckMongoClient
 	lockClient      *dpMongoLock.Lock
+	cfg             *config.Config
 }
 
 // CreateJob creates a new job, with the given id, in the collection, and assigns default values to its attributes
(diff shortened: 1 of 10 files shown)


mongo/job_store.go
```diff
@@ -38,13 +40,37 @@ func (m *JobStore) CreateJob(ctx context.Context, id string) (job models.Job, er
...
+	result := models.Job{}
+
+	// Check that there are no jobs in progress, which started between a calculated "check from" time and now.
+	// The checkFromTime is calculated by subtracting the configured variable "MaxReindexJobRuntime" from now.
+	checkFromTime := time.Now().Add(-1 * m.cfg.MaxReindexJobRuntime)
+	var jobStartTime time.Time
+	for iter.Next(&result) {
+		//if the start time of the job in progress is later than the checkFromTime but earlier than now then a new job should not be created yet.
+		jobStartTime = result.ReindexStarted
+		if jobStartTime.After(checkFromTime) && jobStartTime.Before(time.Now()) {
```
ChrysmOre, line 59: Hmm, would it ever be possible for jobStartTime to not be before now?

mongo/job_store.go
```diff
@@ -38,13 +40,37 @@ func (m *JobStore) CreateJob(ctx context.Context, id string) (job models.Job, er
...
+			log.Event(ctx, "found job in progress", log.Data{"id": result.ID, "state": result.State, "start time": jobStartTime})
+			return models.Job{}, ErrExistingJobInProgress
+		}
+	}
+	defer func() {
+		err := iter.Close()
```
ChrysmOre, line 65: This could be rewritten as one line `if err:= iter.Close; err != ni...`

ChrysmOre (commented):
LGTM, couple of small comments

mongo/job_store.go
cookel2: I think that's a question for my tech lead. I can manually set it to a future time but whether that would actually ever happen I don't know - I'll ask.

mongo/job_store.go
cookel2: nice - done

mongo/job_store.go
cookel2: It looks like it's possible that the jobStartTime could be (at least manually) set to a future time so I'll leave this as is unless you strongly object.

ChrysmOre (approved):
Tested functionality works as expected
Unit tests pass
Component tests pass

PR #23, opened 2021-07-23
Release/0.7.0

cookel2:
# What has changed

These are the changes in this release:

New component tests have been added to test failure scenarios and increase code coverage. The following component tests have been added:

1. Scenario: Job does not exist in the Job Store and a get request returns StatusNotFound
2. Scenario: No Jobs exist in the Job Store and a get request returns an empty list
3. Scenario: The connection to mongo DB is lost and a post request returns an internal server error
4. Scenario: Job does not exist in the Job Store and a put request, to update its number of tasks, returns StatusNotFound
5. Scenario: A put request fails to update the number of tasks because it contains an invalid value of count
6. Scenario: A put request fails to update the number of tasks because it contains a negative value of count
7. Scenario: The connection to mongo DB is lost and a put request returns an internal server error
8. Scenario: The connection to mongo DB is lost and a get request returns an internal server error

The endpoint that allows a POST request to be made, which creates a search reindex job document in mongoDB, has been amended so that the new job will only get created if there are no jobs currently in progress.

If a job is currently in progress then it will contain the following value of state:

"state": "in-progress"

The endpoint now uses a new environment variable, named MAX_REINDEX_JOB_RUNTIME, to calculate a "check from" time.

It checks the reindex_started time of each "in-progress" job to see if it falls between the "check from" time and now. If it does then a 409 conflict error occurs with the following message:

"existing reindex job in progress"

A unit test has been added for the POST request handler.

A few other changes have been made to refactor existing code and increase the coverage of unit tests.

# How to review

Check only expected commits present

# Who can review

Anyone but me


features/steps/jobs_steps.go
```diff
@@ -104,15 +109,30 @@ func (f *JobsFeature) RegisterSteps(ctx *godog.ScenarioContext) {
...
+	ctx.Step(`^I call PUT \/jobs\/{"([^"]*)"}\/number_of_tasks\/{(\d+)} using a valid UUID$`, f.iCallPUTJobsNumber_of_tasksUsingAValidUUID)
+	ctx.Step(`^I call PUT \/jobs\/{id}\/number_of_tasks\/{"([^"]*)"} using the generated id with an invalid count$`, f.iCallPUTJobsidnumber_of_tasksUsingTheGeneratedIdWithAnInvalidCount)
+	ctx.Step(`^I call PUT \/jobs\/{id}\/number_of_tasks\/{"([^"]*)"} using the generated id with a negative count$`, f.iCallPUTJobsidnumber_of_tasksUsingTheGeneratedIdWithANegativeCount)
+	ctx.Step(`^the search reindex api loses its connection to mongo DB$`, f.theSearchReindexApiLosesItsConnectionToMongoDB)
 }
 
 // Reset sets the resources within a specific JobsFeature back to their default values.
-func (f *JobsFeature) Reset() *JobsFeature {
-	f.MongoClient.Database = memongo.RandomDatabase()
+func (f *JobsFeature) Reset(mongoFail bool) *JobsFeature {
```
ChrysmOre, line 124: Why doesn't this function return an error? What happens if one the calls fail and it returns nil?

features/steps/jobs_steps.go
```diff
@@ -452,3 +435,160 @@ func (f *JobsFeature) theJobsShouldBeOrderedByLast_updatedWithTheOldestFirst() e
...
+	_, err := uuid.FromString(id)
+	if err != nil {
+		fmt.Println("Got uuid: " + id)
+		return err
+	}
+
+	// call GET /jobs/{id}
+	err = f.ApiFeature.IGet("/jobs/" + id)
+	if err != nil {
+		os.Exit(1)
```
ChrysmOre, line 572: I think it would be helpful to have some kind of log to let us know why it's failing here

features/steps/jobs_steps.go
```diff
@@ -452,3 +435,160 @@ func (f *JobsFeature) theJobsShouldBeOrderedByLast_updatedWithTheOldestFirst() e
...
+}
+
+// PutNumberOfTasks is a utility function that is used for calling the PUT /jobs/{id}/number_of_tasks/{count}
+// It checks that the id string is a valid UUID before calling the endpoint.
+func (f *JobsFeature) PutNumberOfTasks(countStr string) error {
+
+	var emptyBody = godog.DocString{}
+	_, err := uuid.FromString(id)
+	if err != nil {
+		fmt.Println("Got uuid: " + id)
```
ChrysmOre, line 584: Is this log kept here intentionally or is it left over debug? Actually not sure it makes sense to say it successfully got the ID if it a failed?

features/steps/jobs_steps.go
ChrysmOre: Actually why is it exiting rather than returning the error?

features/steps/jobs_steps.go
```diff
@@ -452,3 +435,160 @@ func (f *JobsFeature) theJobsShouldBeOrderedByLast_updatedWithTheOldestFirst() e
...
+		return err
+	}
+
+	// call PUT /jobs/{id}/number_of_tasks/{count}
+	err = f.ApiFeature.IPut("/jobs/"+id+"/number_of_tasks/"+countStr, &emptyBody)
+	if err != nil {
+		os.Exit(1)
```
ChrysmOre, line 591: Again here exiting without any context rather than returning the error?

features/steps/jobs_steps.go
cookel2: Yes you are right to suggest that this should be returning an error I think. The function was originally copied from somewhere else so I think it usually returns just the component to test (e.g. the Reset() function in the dataset_component.go class in dp-dataset-api does). However, as I've added more functionality to this one then it's a good point.

features/steps/jobs_steps.go
cookel2: I think there's a historical reason for that but I'm not sure what it was. In refactoring this code copied this line from when the tests had first started being written.

features/steps/jobs_steps.go
cookel2: Again, this was copied from historical code, and it's a very good point. It needs to be removed.

features/steps/jobs_steps.go
cookel2: yes, I think those both need changing as you suggest

features/steps/jobs_steps.go
cookel2: done

features/steps/jobs_steps.go
cookel2: done

features/steps/jobs_steps.go
cookel2: done

cookel2:
closing this PR - I've made all the changes in a new branch named feature/component-test-error-handling

PR #24, opened 2021-07-26
Feature/component test error handling

cookel2:
# What has changed

These changes have been made, to the jobs_steps.go, as a result of feedback on a PR for release 0.7.0 (now closed):

- The Reset function now returns an error. Also, it will now only try to set the Config and initialise the MongoClient if those things have not already been done.

- a couple of functions that were just exiting now return an error instead.

- a couple of Println statements left over from debugging have now been removed.

And logging has been added to main_test.go wherever it exits.

# How to review

The existing functionality can be tested in develop using these commands:

Post a default job: curl -X POST http://localhost:12150/jobs

Get all the jobs: curl -s http://localhost:12150/jobs | jq

Get a specific job (e.g. with id = d2fddaf5-d812-41a3-bd73-ec60eefbb813): curl -s http://localhost:12150/jobs/d2fddaf5-d812-41a3-bd73-ec60eefbb813 | jq

Update the number of tasks for a specific job: curl -X PUT http://localhost:12150/jobs/9f1469a7-97fa-4344-bbea-28906a61b1c1/number_of_tasks/7

Check the health of the dp-search-reindex-api: curl -s http://localhost:12150/health | jq

# Who can review

Anyone but me

### DIFF
--- a/features/steps/jobs_steps.go
+++ b/features/steps/jobs_steps.go
@@ -4,6 +4,7 @@ package steps
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"os"
@@ -131,18 +132,18 @@ func (f *JobsFeature) Reset(mongoFail bool) error {
 	if f.Config == nil {
 		cfg, err := config.Get()
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to get config: %w", err)
 		}
 		f.Config = cfg
 	}
 	if f.MongoClient == nil {
 		err = f.MongoClient.Init(ctx, f.Config)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to initialise Mongo client: %w", err)
 		}
 	}
 
-	return err
+	return nil
 }
 
 // Close stops the *service.Service, which is pointed to from within the specific JobsFeature, from running.
--- a/features/steps/jobs_steps.go
+++ b/features/steps/jobs_steps.go
@@ -57,7 +57,7 @@ func NewJobsFeature(mongoFeature *componentTest.MongoFeature) (*JobsFeature, err
 	svcErrors := make(chan error, 1)
 	cfg, err := config.Get()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to get config: %w", err)
 	}
 	mongodb := &mongo.JobStore{
 		Collection: jobsCol,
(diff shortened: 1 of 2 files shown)


features/steps/jobs_steps.go
```diff
@@ -121,22 +120,29 @@ func (f *JobsFeature) RegisterSteps(ctx *godog.ScenarioContext) {
...
 		f.MongoClient.Database = memongo.RandomDatabase()
 	}
 	ctx := context.Background()
-	cfg, err := config.Get()
-	if err != nil {
-		return nil
+	if f.Config == nil {
+		cfg, err := config.Get()
+		if err != nil {
+			return err
```
ChrysmOre, line 134: So as I talked about in my presentation about errors I think now's a good a time as any to start encouraging people.. Pretty much all the time instead of just returning an error it's better to add a small amount of context describing what the action was that just failed. I.E. here  `return err` would become `return fmt.Errorf("failed to get config: %w", err)` and below would become `return fmt.Errorf("failed to initialise Mongo client: %w"). If we do this each and every time we return an error we end up with an error which is really informative and helps us debug much more efficiently.

features/steps/jobs_steps.go
```diff
@@ -121,22 +120,29 @@ func (f *JobsFeature) RegisterSteps(ctx *godog.ScenarioContext) {
...
+		}
+		f.Config = cfg
 	}
-	err = f.MongoClient.Init(ctx, cfg)
-	if err != nil {
-		return nil
+	if f.MongoClient == nil {
+		err = f.MongoClient.Init(ctx, f.Config)
```
ChrysmOre, line 139: This doesn't look right. If `f.MongoClient == nil` this would panic when trying to call a function on it?

ChrysmOre (changes requested):

features/steps/jobs_steps.go
```diff
@@ -121,22 +120,29 @@ func (f *JobsFeature) RegisterSteps(ctx *godog.ScenarioContext) {
...
+		if err != nil {
+			return err
+		}
 	}
-	return f
+
+	return err
```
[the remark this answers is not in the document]
cookel2: ok, done

main_test.go
```diff
@@ -24,19 +27,27 @@ type ComponentTest struct {
 func (f *ComponentTest) InitializeScenario(ctx *godog.ScenarioContext) {
 	authorizationFeature := componentTest.NewAuthorizationFeature()
 	jobsFeature, err := steps.NewJobsFeature(f.MongoFeature)
+
+	ctxBackground := context.Background()
```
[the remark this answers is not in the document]
cookel2: done, I've also renamed it to ctx to be more consistent with the rest of the repo (and renamed the other one to godogCtx)

features/steps/jobs_steps.go
cookel2: done, thanks for explaining how to do this

features/steps/jobs_steps.go
cookel2: I added the 'if f.MongoClient == nil' statement because otherwise the component tests produce the following error: failed to initialise Mongo client: session already exists

ChrysmOre: Still, you can't call a function on a nil pointer - it will panic. What it looks like is that you need to remove the re-initialisation from this function altogether.

cookel2: I see what you mean, thanks, I've removed that now

ChrysmOre (approved):
Much better! Thanks

PR #26, opened 2021-08-12
Feature/add pagination to jobs

cookel2:
# What has changed

As an API user
I want to be able to use limit and offset parameters on list endpoints
so that I can page through jobs

These are the changes:

The endpoint that allows a GET request to be made, which lists search reindex job documents from mongoDB, has been amended so that the list returned will also be determined by pagination functionality i.e. offset and limit query parameters.

The pagination functionality should meet the API standards defined here: https://github.com/ONSdigital/dp/blob/main/standards/API_STANDARDS.md

If the offset or limit query parameters fail validation then a 400 (Bad Request) error occurs.

Also, as well as returning the list of jobs, the endpoint now returns the following values (as defined by the swagger spec):
- count (The number of Job resources returned in this response.)
- limit (The max number of Job resources that should be returned in this response.)
- offset (The number of Job resources into the full list, that is the 0-based index, that this particular response is starting at.)
- total_count (How many total items there may be - the total number of jobs in mongoDB.)

Default values have also been added, for the limit and offset query parameters, and a default maximum limit. These are configurable and given by the following environment variables:

DEFAULT_OFFSET (currently set to 0)
DEFAULT_LIMIT (currently set to 20)
DEFAULT_MAXIMUM_LIMIT (currently set to 1000)

Five component tests have been added to the getting_jobs feature, as follows:

1. Scenario: Six jobs exist and a get request with offset and limit correctly returns four
2. Scenario: Three jobs exist and a get request with negative offset returns an error
3. Scenario: Three jobs exist and a get request with offset greater than three returns an error
4. Scenario: Three jobs exist and a get request with negative limit returns an error
5. Scenario: Three jobs exist and a get request with limit greater than the maximum returns an error

Some unit test have also been added.

# How to review

If testing this locally then first use dp-compose to run mongoDB in a docker container. Then hit the POST endpoint a few times to create some jobs:
http://localhost:25700/jobs

The changes can then be tested using this GET URL or similar: 
http://localhost:25700/jobs?offset=3&limit=100

# Who can review

Anyone but me

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -1,4 +1,4 @@
-# The nancy command has found a vulnerability, which is an out of date version of GoGo Protobuf. However, this cannot be updated due to the project also having an indirect dependency of something called 
+# The nancy command has found a vulnerability, which is an out of date version of GoGo Protobuf. However, this cannot be updated due to the project also having an indirect dependency of something called
 # scram; the latest version of scram has a new path in github, which also needs to be changed (instead of http://github.com/xdg/scram it is now http://github.com/xdg-go/scram I believe). The problem is 
 # that if I manually amend this path and version in the go.sum file then it reverts back when I try to commit and push the code (so it won't let me update GoGo Protobuf either). So I need to update the
 # dependency that is actually using scram directly. That dependency is actually mongo-driver, which itself is a dependency in dp-component-test, which is a dependency in this project. But on further
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"github.com/ONSdigital/dp-search-reindex-api/pagination"
 	"math"
 	"net/http"
 	"strconv"
@@ -14,12 +15,22 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
-// NewID generates a random uuid and returns it as a string.
-var NewID = func() string {
-	return uuid.NewV4().String()
-}
+var (
+	// NewID generates a random uuid and returns it as a string.
+	NewID = func() string {
+		return uuid.NewV4().String()
+	}
 
-var serverErrorMessage = "internal server error"
+	serverErrorMessage = "internal server error"
+
+	// errors that should return a 400 status
+	badRequest = map[error]bool{
+		pagination.ErrInvalidLimitParameter:  true,
+		pagination.ErrInvalidOffsetParameter: true,
+		pagination.ErrLimitOverMax:           true,
+		pagination.ErrOffsetOverTotalCount:   true,
+	}
+)
 
 // CreateJobHandler returns a function that generates a new Job resource containing default values in its fields.
 func (api *JobStoreAPI) CreateJobHandler(ctx context.Context) http.HandlerFunc {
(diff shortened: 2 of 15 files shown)


.nancy-ignore
```diff
@@ -1,4 +1,4 @@
-# The nancy command has found a vulnerability, which is an out of date version of GoGo Protobuf. However, this cannot be updated due to the project also having an indirect dependency of something called 
+# The nancy command has found a vulnerability, which is an out of date version of GoGo Protobuf. However, this cannot be updated due to the project also having an indirect dependency of something called
```
bpathak-ons, line 1: Error with "make audit" needs to be fixed.... 
go list -json -m all | nancy sleuth
/bin/sh: nancy: command not found
make: *** [audit] Error 127

api/jobs.go
```diff
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"github.com/ONSdigital/dp-search-reindex-api/pagination"
```
bpathak-ons, line 6: explicit import not needed, I think

mongo/job_store.go
```diff
@@ -167,7 +168,18 @@ func (m *JobStore) GetJobs(ctx context.Context) (models.Jobs, error) {
 		return results, err
 	}
 
+	paginator := pagination.NewPaginator(m.cfg.DefaultLimit, m.cfg.DefaultOffset, m.cfg.DefaultMaxLimit)
+	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam, numJobs)
+	if err != nil {
+		return results, err
```
bpathak-ons, line 174: should we capture error details here using errorf?

pagination/pagination.go
```diff
@@ -0,0 +1,75 @@
...
+// ValidatePaginationParameters returns pagination related values based on the given request
+func (p *Paginator) ValidatePaginationParameters(offsetParameter string, limitParameter string, totalCount int) (offset int, limit int, err error) {
+
+	offset = p.DefaultOffset
+	limit = p.DefaultLimit
+
+	if offsetParameter != "" {
+		offset, err = strconv.Atoi(offsetParameter)
+		if err != nil || offset < 0 {
+			return 0, 0, ErrInvalidOffsetParameter
```
bpathak-ons, line 55: possibly errors for logging purpose can be captured using errorf

pagination/pagination.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	if offset > totalCount {
+		return 0, 0, ErrOffsetOverTotalCount
```
bpathak-ons, line 60: see above for capturing error logging

.nancy-ignore
cookel2: This error only occurs locally, I think it works ok if nancy is installed, the audit checks pass in concourse.

mongo/job_store.go
cookel2: My understanding is that we should either log errors or return them but there's no need to do both. The details of these errors get captured in the handler and dealt with by sending the appropriate error code and message back in the response.

pagination/pagination.go
cookel2: This is the way that it's done in dp-collection-api (in the develop branch) so I would keep it the same as that - also I don't think we necessarily need to log errors if we are returning them.

pagination/pagination.go
cookel2: same as above

api/jobs.go
cookel2: it does seem to be needed but it's not in the right place so good spot - I will change it

cookel2: done

PR #31, opened 2021-09-20
Add post task endpoint

cookel2:
# What has changed

As an internal service
I want to create a single task against a job
so that the task can be tracked separately from the job

These are the changes:

An endpoint has been added, which allows a POST request to be made, to create or overwrite (if it already exists) a named task for a given job. The task will be a document in the tasks collection of the search database in MongoDB. The relevant task is returned as the response.

The endpoint is /jobs/{job id}/tasks

{
  "task_name": "api service name",
  "number_of_documents": 29
}

Example (to create a task for the job id 883c81fd-726d-4ea3-9db8-7e7c781a01cc):
http://localhost:25700/jobs/883c81fd-726d-4ea3-9db8-7e7c781a01cc/tasks

Request body:

{
  "task_name": "dp-hierarchy-api",
  "number_of_documents": 18
}

Response:

{
  "job_id": "883c81fd-726d-4ea3-9db8-7e7c781a01cc",
  "last_updated": "2021-03-18T09:03:41+0000",
  "links": {
    "self": "http://localhost:12150/jobs/abc123/tasks/dp-dataset-api",
    "job": "http://localhost:12150/jobs/abc123"
  },
  "number_of_documents": 18,
  "task_name": "dp-hierarchy-api"
}

NB. The new endpoint requires authorisation of type 'Bearer Token'. The token it needs is the SERVICE_AUTH_TOKEN so this must be set appropriately in whatever environment is being used. This also means that Zebedee (and Vault, which Zebedee has a dependency on) must now also be running in the same environment as this service. The README has been updated to reflect this.

A new feature has been added to the component tests, named posting_a_task, which contains 9 component tests as follows:

1. Scenario: Task is created successfully
2. Scenario: The connection to mongo DB is lost and a post request returns an internal server error
3. Scenario: Task is updated successfully
4. Scenario: Request body cannot be read returns an internal server error
5. Scenario: Job does not exist and an attempt to create a task for it returns a not found error    
6. Scenario: No authorisation header set returns a bad request error
7. Scenario: Invalid service auth token returns unauthorised error
8. Scenario: Valid user token returns bad request error
9. Scenario: Invalid user token returns bad request error

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

Then hit the POST endpoint to create at job. Take note of the job id that it generates:
http://localhost:25700/jobs

The changes can then be tested locally using the POST URL e.g for a job with id 1b9a6881-8bf3-4729-8c59-62901c4688b8: 

{
  "task_name": "dp-hierarchy-api",
  "number_of_documents": 18
}

# Who can review

Anyone but me


api/tasks.go
```diff
@@ -0,0 +1,56 @@
...
+		taskToCreate := &models.TaskToCreate{}
+		if err := ReadJSONBody(ctx, req.Body, taskToCreate); err != nil {
+			log.Error(ctx, "reading request body failed", err)
+			http.Error(w, serverErrorMessage, http.StatusBadRequest)
+			return
+		}
+
+		newTask, err := api.jobStore.CreateTask(ctx, jobID, taskToCreate.TaskName, taskToCreate.NumberOfDocuments)
+		if err != nil {
+			log.Error(ctx, "creating and storing a task failed", err)
```
daiLlew, line 31: Consider adding the job ID to the log data as it is useful context that will aid investigating errors.

models/job.go
```diff
@@ -26,15 +31,22 @@ type JobLinks struct {
...
 
 // NewJob returns a new Job resource that it creates and populates with default values.
-func NewJob(id string) Job {
+func NewJob(id string) (Job, error) {
 	zeroTime := time.Time{}.UTC()
+	cfg, err := config.Get()
+	if err != nil {
+		return Job{}, fmt.Errorf("%s: %w", errors.New("unable to retrieve service configuration"), err)
+	}
+	urlBuilder := url.NewBuilder("http://" + cfg.BindAddr)
```
daiLlew, line 40: It might be worth specifying the expected format for `BindAddr` in the readme  i.e. it doesn't require a scheme.

models/job.go
```diff
@@ -44,5 +56,5 @@ func NewJob(id string) Job {
 		State:                        "created",
 		TotalSearchDocuments:         0,
 		TotalInsertedSearchDocuments: 0,
-	}
+	}, err
```
daiLlew, line 59: Consider returning `nil` here instead - I think its clearer.

models/task.go
```diff
@@ -0,0 +1,47 @@
...
+
+// TaskLinks is a type that contains links to the endpoints for returning a specific task (self), and the job that it is part of (job), respectively.
+type TaskLinks struct {
+	Self string `json:"self"`
+	Job  string `json:"job"`
+}
+
+// NewTask returns a new Task resource that it creates and populates with default values.
+func NewTask(jobID string, taskName string, numDocuments int) (Task, error) {
+	cfg, err := config.Get()
```
daiLlew, line 29: Consider providing  the bind address as a param. The config is only required to get the bind address I think it reasonable to provide that as a single param. Also if it is a param then you will no longer need to get the config and wont need to return an error from this methods.

models/task_to_create.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+// TaskToCreate is a type that contains the details required for creating a Task type.
+type TaskToCreate struct {
```
daiLlew, line 4: Is there a reason this is split into its own file? Could it live with the other task models?

mongo/job_store.go
```diff
@@ -92,6 +97,45 @@ func (m *JobStore) CreateJob(ctx context.Context, id string) (job models.Job, er
...
+	jobToFind.ID = jobID
+	err := s.DB(m.Database).C(m.JobsCollection).Find(bson.M{"_id": jobID}).One(&jobToFind)
+	if err != nil {
+		log.Error(ctx, "error finding job for task", err)
+		if err == mgo.ErrNotFound {
+			return models.Task{}, ErrJobNotFound
+		} else {
+			return models.Task{}, fmt.Errorf("an unexpected error has occurred: %w", err)
+		}
+	}
```
daiLlew, line 123: This seems to be common across multiple updates. Consider extracting this out into a separate method.

mongo/job_store.go
```diff
@@ -239,6 +284,26 @@ func (m *JobStore) UpdateJob(updates bson.M, s *mgo.Session, id string) error {
...
 }
 
+// UpsertTask creates a new task document or overwrites an existing one
+func (m *JobStore) UpsertTask(jobID, taskName string, task models.Task) (err error) {
+	s := m.Session.Copy()
+	defer s.Close()
+
+	selector := bson.M{
+		"task_name": taskName,
+		"job_id":    jobID,
```
daiLlew, line 294: Do we want to check job ID is not empty?

mongo/job_store.go
```diff
@@ -239,6 +284,26 @@ func (m *JobStore) UpdateJob(updates bson.M, s *mgo.Session, id string) error {
 	return nil
 }
 
+// UpsertTask creates a new task document or overwrites an existing one
+func (m *JobStore) UpsertTask(jobID, taskName string, task models.Task) (err error) {
```
daiLlew, line 288: Parentheses not required for methods/funcs signatures returning single value.

api/tasks.go
cookel2: good idea - done

models/job.go
cookel2: done

models/job.go
cookel2: yes I agree - done

mongo/job_store.go
cookel2: good idea - done

models/task.go
cookel2: yes, good idea, I wish I had thought of it

models/task_to_create.go
cookel2: I thought it was the convention but maybe not - I've moved it now

mongo/job_store.go
cookel2: The CreateTask function is the only one that calls this method. That function already checks to see if the job id is empty so it doesn't really need to be checked again here unless it's going to be called by something else - it isn't currently though.

mongo/job_store.go
cookel2: good point - changed

daiLlew (approved):

PR #32, opened 2021-09-27
Improve the Validation on the CreateTask endpoint

cookel2:
### What

Correct the message for a bad request
Check that the task name is not an empty string

### How to review

The endpoint is /jobs/{job id}/tasks

{ “task_name”: <api service>, “number_of_documents”:5}

### Who can review

@nshumoogum

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -11,6 +11,8 @@ import (
 	"github.com/gorilla/mux"
 )
 
+var invalidBodyErrorMessage = "invalid request body"
+
 // CreateTaskHandler returns a function that generates a new TaskName resource containing default values in its fields.
 func (api *JobStoreAPI) CreateTaskHandler(ctx context.Context) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
@@ -22,7 +24,7 @@ func (api *JobStoreAPI) CreateTaskHandler(ctx context.Context) http.HandlerFunc 
 		taskToCreate := &models.TaskToCreate{}
 		if err := ReadJSONBody(ctx, req.Body, taskToCreate); err != nil {
 			log.Error(ctx, "reading request body failed", err)
-			http.Error(w, serverErrorMessage, http.StatusBadRequest)
+			http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
 			return
 		}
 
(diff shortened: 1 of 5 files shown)


mongo/job_store.go
```diff
@@ -98,14 +98,18 @@ func (m *JobStore) CreateJob(ctx context.Context, id string) (models.Job, error)
...
-	log.Info(ctx, "creating task in mongo DB", log.Data{"jobID": jobID, "nameOfApi": nameOfApi, "numDocuments": numDocuments})
+func (m *JobStore) CreateTask(ctx context.Context, jobID string, taskName string, numDocuments int) (models.Task, error) {
+	log.Info(ctx, "creating task in mongo DB", log.Data{"jobID": jobID, "taskName": taskName, "numDocuments": numDocuments})
 
 	// If an empty job id was passed in, return an error with a message.
 	if jobID == "" {
 		return models.Task{}, ErrEmptyIDProvided
 	}
 
+	if taskName == "" {
```
nshumoogum, line 109: This would be better handled inside the handler, not the mongo store.

So inside `CreateTaskHandler` handler:

```
    if err := task.Validate(); err != nil {
        log.Error(ctx, "CreateTask endpoint: Invalid request body", err)
        http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
    }
```

where in models/task.go:

```
func (task *Task) Validate() error {
	if task.TaskName == {
        	return errs.ErrEmptyTaskNameProvided
        }
}
```

where errs is the apierrors package and inside apierrors.go add `var ErrEmptyTaskNameProvided = errors.New("task name must not be an empty string")`

mongo/errors.go
```diff
@@ -8,6 +8,7 @@ import (
 var (
 	ErrJobNotFound           = errors.New("the job id could not be found in the jobs collection")
 	ErrEmptyIDProvided       = errors.New("id must not be an empty string")
+	ErrEmptyTaskNameProvided = errors.New("task name must not be an empty string")
```
nshumoogum, line 11: As shown in previous comment, this will live inside apierrors package

api/tasks.go
```diff
@@ -31,6 +33,8 @@ func (api *JobStoreAPI) CreateTaskHandler(ctx context.Context) http.HandlerFunc
 			log.Error(ctx, "creating and storing a task failed", err, log.Data{"job id": jobID})
 			if err == mongo.ErrJobNotFound {
 				http.Error(w, "Failed to find job that has the specified id", http.StatusNotFound)
+			} else if err == mongo.ErrEmptyTaskNameProvided {
```
nshumoogum, line 36: This else if statement wont be needed anymore

nshumoogum (changes requested):

nshumoogum (approved):

PR #34, opened 2021-10-01
Use version 0.6.0 of dp-component-test and refactor tests to use it

cookel2:
# What has changed

This service now uses version 0.6.0 of the dp-component-test library. And it has been refactored to make use of the following six new scenario steps that the library provides:
1. I use a service auth token
2. I use an X Florence user token
3. zebedee does not recognise the service auth token
4. zebedee recognises the service auth token as valid 
5. zebedee does not recognise the user token
6. zebedee recognises the user token as valid

# How to review

Make sure that the component tests all still pass when you run this command:

go test -component

NB. 114 steps should pass

# Who can review

Anyone but me


rafahop (approved):
Looks good to me 👍

PR #38, opened 2021-10-14
Limit task names

cookel2:
# What has changed

Limit the task names to enumerated list

As a Developer
I want to prevent creation of unknown tasks
so that the search reindex api can correctly track reindexing of search

These are the changes:

The existing endpoint, which allows a POST request to be made, to create a named task for a given job, has been amended. This is so that the chosen task name will be restricted to those that are listed in an enum. The enum will contain the following task name values:
- dataset-api
- zebedee

The amended POST endpoint is /jobs/{job id}/tasks

{ “task_name”: VALUE-FROM-ENUM, “number_of_documents”: NUMERIC-VALUE}

Example of a valid request body:
{ “task_name”: "zebedee", “number_of_documents”:12}

Example of an invalid request body:
{ “task_name”: "babbage", “number_of_documents”:5}

A new scenario has been added to the posting_a_task feature of the component tests. It is:
Scenario: Invalid task name returns bad request error

NB. The following existing scenario already tests for success when the task name is valid:
Scenario: Task is created successfully
    
Unit tests have also been added.

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

Then hit the POST endpoint to create at job. Take note of the job id that it generates:
http://localhost:25700/jobs

The changes can then be tested locally using the POST URL e.g for a job with id 1b9a6881-8bf3-4729-8c59-62901c4688b8: 

{
  "task_name": "dataset-api",
  "number_of_documents": 18
}

# Who can review

Anyone but me

### DIFF
--- a/api/tasks.go
+++ b/api/tasks.go
@@ -12,52 +12,50 @@ import (
 var invalidBodyErrorMessage = "invalid request body"
 
 // CreateTaskHandler returns a function that generates a new TaskName resource containing default values in its fields.
-func (api *API) CreateTaskHandler() http.HandlerFunc {
-	return func(w http.ResponseWriter, req *http.Request) {
-		ctx := req.Context()
-		vars := mux.Vars(req)
-		jobID := vars["id"]
+func (api *API) CreateTaskHandler(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	vars := mux.Vars(req)
+	jobID := vars["id"]
 
-		// Unmarshal task to create and validate it
-		taskToCreate := &models.TaskToCreate{}
-		if err := ReadJSONBody(ctx, req.Body, taskToCreate); err != nil {
-			log.Error(ctx, "reading request body failed", err)
-			http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
-			return
-		}
+	// Unmarshal task to create and validate it
+	taskToCreate := &models.TaskToCreate{}
+	if err := ReadJSONBody(ctx, req.Body, taskToCreate); err != nil {
+		log.Error(ctx, "reading request body failed", err)
+		http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
+		return
+	}
 
-		if err := taskToCreate.Validate(api.taskNames); err != nil {
-			log.Error(ctx, "CreateTask endpoint: Invalid request body", err)
-			http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
-			return
-		}
+	if err := taskToCreate.Validate(api.taskNames); err != nil {
+		log.Error(ctx, "CreateTask endpoint: Invalid request body", err)
+		http.Error(w, invalidBodyErrorMessage, http.StatusBadRequest)
+		return
+	}
 
-		newTask, err := api.dataStore.CreateTask(ctx, jobID, taskToCreate.TaskName, taskToCreate.NumberOfDocuments)
-		if err != nil {
-			log.Error(ctx, "creating and storing a task failed", err, log.Data{"job id": jobID})
-			if err == mongo.ErrJobNotFound {
-				http.Error(w, "Failed to find job that has the specified id", http.StatusNotFound)
-			} else {
-				http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			}
-			return
-		}
-
-		jsonResponse, err := json.Marshal(newTask)
-		if err != nil {
-			log.Error(ctx, "marshalling response failed", err)
+	newTask, err := api.dataStore.CreateTask(ctx, jobID, taskToCreate.TaskName, taskToCreate.NumberOfDocuments)
+	if err != nil {
+		log.Error(ctx, "creating and storing a task failed", err, log.Data{"job id": jobID})
+		if err == mongo.ErrJobNotFound {
+			http.Error(w, "Failed to find job that has the specified id", http.StatusNotFound)
+		} else {
 			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
 		}
+		return
+	}
 
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusCreated)
-		_, err = w.Write(jsonResponse)
-		if err != nil {
-			log.Error(ctx, "writing response failed", err)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
-		}
+	jsonResponse, err := json.Marshal(newTask)
+	if err != nil {
+		log.Error(ctx, "marshalling response failed", err)
+		http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusCreated)
+	_, err = w.Write(jsonResponse)
+	if err != nil {
+		log.Error(ctx, "writing response failed", err)
+		http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+		return
 	}
 }
 
--- a/api/tasks_test.go
+++ b/api/tasks_test.go
@@ -62,7 +62,7 @@ func TestCreateTaskHandler(t *testing.T) {
 
 	Convey("Given an API that can create valid search reindex tasks and store their details in a Data Store", t, func() {
 		apiInstance := api.Setup(ctx, mux.NewRouter(), dataStorerMock, &apiMock.AuthHandlerMock{}, taskNames)
-		createTaskHandler := apiInstance.CreateTaskHandler()
+		createTaskHandler := apiInstance.CreateTaskHandler
 
 		Convey("When a new reindex task is created and stored", func() {
 			req := httptest.NewRequest("POST", fmt.Sprintf("http://localhost:25700/jobs/%s/tasks", validJobID1), bytes.NewBufferString(
@@ -71,7 +71,7 @@ func TestCreateTaskHandler(t *testing.T) {
 			req.Header.Set("Authorization", validServiceAuthToken)
 			resp := httptest.NewRecorder()
 
-			createTaskHandler.ServeHTTP(resp, req)
+			createTaskHandler(resp, req)
 
 			Convey("Then the newly created search reindex task is returned with status code 201", func() {
 				So(resp.Code, ShouldEqual, http.StatusCreated)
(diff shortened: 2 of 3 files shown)


models/task.go
```diff
@@ -22,6 +26,24 @@ type TaskLinks struct {
 	Job  string `json:"job"`
 }
 
+// TaskName - iota enum of possible task names
+type TaskName int
+
+// ParseTaskName returns a TaskName from its string representation
+func ParseTaskName(taskNameStr string) (TaskName, error) {
+	cfg, err := config.Get()
```
nshumoogum, line 34: Never reinstantiate the config while the app is processing (this should only be done once at start up and the values passed to where they are needed in the app).

The reason is that, someone can make changes to environment variables on disk while this app is running. In this case your code will bring them in when this function gets called instead of using the configuration defined on startup.

This would put the app in a weird state where some functions it has behave in the expected way due to configurations inherited from start up while here it may behave differently due to picking up new values retrieved at this point in time.

models/task.go
```diff
@@ -22,6 +26,24 @@ type TaskLinks struct {
...
+	if err != nil {
+		return -1, fmt.Errorf("%s: %w", errors.New("unable to retrieve service configuration"), err)
+	}
+	taskNameValues := strings.Split(cfg.TaskNameValues, ",")
```
nshumoogum, line 38: This can be done once on startup, instead of calculating this logic everytime a request comes in - this will increase performance

Also it would be better to create a map instead of an array, so we can just lookup the taskname in map instead of using a for loop when the request comes in - again increases performance.

models/task.go
```diff
@@ -22,6 +26,24 @@ type TaskLinks struct {
 	Job  string `json:"job"`
 }
 
+// TaskName - iota enum of possible task names
+type TaskName int
+
+// ParseTaskName returns a TaskName from its string representation
+func ParseTaskName(taskNameStr string) (TaskName, error) {
```
nshumoogum, line 33: I don't see the need for this function returning `TaskName` as it is thrown away by the calling method on line 76 inside Validate method.

README.md
```diff
@@ -24,7 +24,7 @@ In the zebedee directory run `./run.sh` to run Zebedee
 
 | Environment variable         | Default               | Description
 | ---------------------------- | --------------------- | -----------
-| BIND_ADDR                    | localhost:25700       | The host and port to bind to (The http:// scheme prefix is added programmatically)
+| BIND_ADDR                    | localhost:25700       | The host and port to bind to (The http:// scheme prefix is added programmatically)
```
ChrysmOre, line 27: Pedantic but extra whitespace at the end of line

api/tasks_test.go
```diff
@@ -0,0 +1,185 @@
...
+// Create Task Payload
+var createTaskPayloadFmt = `{
+	"task_name": "%s",
+	"number_of_documents": 5
+}`
+
+func ExpectedTask(jobID string, lastUpdated time.Time, numberOfDocuments int, taskName string) (models.Task, error) {
+	cfg, err := config.Get()
+	if err != nil {
+		return models.Task{}, fmt.Errorf("%s: %w", errors.New("unable to retrieve service configuration"), err)
```
ChrysmOre, line 171: There's no need to use errors.New() here, as it's being read as it's string interpretation by %s. Simply:
```
 fmt.Errorf("%s: %w", "unable to retrieve service configuration", err)
 ```
 or even
 ```
  fmt.Errorf("unable to retrieve service configuration: %w", err)
  ```
  will produce the same result

File: api/tasks_test.go [hunk 6]
```diff
@@ -0,0 +1,185 @@
...
+	}
+	urlBuilder := url.NewBuilder("http://" + cfg.BindAddr)
+	job := urlBuilder.BuildJobURL(jobID)
+	self := urlBuilder.BuildJobTaskURL(jobID, taskName)
+	return models.Task{
+		JobID:       jobID,
+		LastUpdated: lastUpdated,
+		Links: &models.TaskLinks{
+			Self: self,
+			Job:  job},
```
ChrysmOre, line 181: It's better to close braces on the same column they were open i.e:
```
		Links: &models.TaskLinks{
			Self: self,
			Job:  job,
		},
```

features/posting_a_task.feature
```diff
@@ -119,6 +119,18 @@ Feature: Posting a job
...
+
+    Given I use a service auth token "validServiceAuthToken"
+    And zebedee recognises the service auth token as valid
+    And I have generated a job in the Job Store
+    And I call POST /jobs/{id}/tasks using the generated id
+    """
+    { "task_name": "florence", "number_of_documents": 29 }
+    """
+    Then the HTTP status code should be "400"
+
```
ChrysmOre, line 136: Looks like something strange at the end of the file (stray space?)

models/task.go
```diff
@@ -22,6 +26,24 @@ type TaskLinks struct {
...
+	cfg, err := config.Get()
+	if err != nil {
+		return -1, fmt.Errorf("%s: %w", errors.New("unable to retrieve service configuration"), err)
```
ChrysmOre, line 36: Again this errors.New() is redundant.

api/tasks_test.go
cookel2: thanks, that's helpful - done

features/posting_a_task.feature
cookel2: I normally put a new line at the end of each file (or sometimes the formatter does it for me). Should I not do that here?

ChrysmOre: The newline is fine but it looked like there was a space on the newline or something?

api/api_test.go
```diff
@@ -2,20 +2,29 @@ package api_test
...
 	"github.com/ONSdigital/dp-search-reindex-api/api"
 	"github.com/ONSdigital/dp-search-reindex-api/api/mock"
+	"github.com/ONSdigital/dp-search-reindex-api/config"
 	"github.com/gorilla/mux"
 	. "github.com/smartystreets/goconvey/convey"
 	"net/http/httptest"
 )
 
 func TestSetup(t *testing.T) {
+	cfg, _ := config.Get()
```
ChrysmOre, line 17: Sorry didn't spot this before, but there's really never a good reason to skip checking an error. Can only cause problems down hassle down the line.

api/api_test.go
```diff
@@ -26,10 +35,18 @@ func TestSetup(t *testing.T) {
 }
 
 func TestClose(t *testing.T) {
+	cfg, _ := config.Get()
```
ChrysmOre, line 38: errcheck here too

api/jobs_test.go
```diff
@@ -41,7 +41,15 @@ var ctx = context.Background()
 func TestCreateJobHandler(t *testing.T) {
 	t.Parallel()
 
-	jobsCollectionMock := &apiMock.JobStorerMock{
+	cfg, _ := config.Get()
```
ChrysmOre, line 44: and here

api/jobs_test.go
```diff
@@ -132,8 +140,18 @@ func TestCreateJobHandler(t *testing.T) {
 
 func TestGetJobHandler(t *testing.T) {
 	t.Parallel()
+
+	cfg, _ := config.Get()
```
ChrysmOre, line 144: errcheck

api/jobs_test.go
```diff
@@ -231,8 +249,17 @@ func TestGetJobHandler(t *testing.T) {
 }
 
 func TestGetJobsHandler(t *testing.T) {
+	cfg, _ := config.Get()
```
ChrysmOre, line 252: Looks like several more instances of skipping this error check throughout

ChrysmOre (changes requested):
Looks like there's an error check being skipped in multiple places throughout the tests, generally there's never a good reason and can only cause issues further down the line. I also replied to the comment about the newline.. something is going on, not quite sure but it looks like the newline at the end of the file isn't clean somehow

api/jobs_test.go
```diff
@@ -41,7 +41,18 @@ var ctx = context.Background()
 func TestCreateJobHandler(t *testing.T) {
 	t.Parallel()
 
-	jobsCollectionMock := &apiMock.JobStorerMock{
+	cfg, err := config.Get()
```
nshumoogum, line 44: So these are repetitive blocks of code (lines 44 to 54) in each test, this could be moved into a separate function that each Test calls before running any assertions. Also you are overwriting the t variable name which is set to *testing.T, better to use a different variable in for loop (line 52).

But all of the above wont matter as I don't like the idea of using the configuration values to determine whether the code in create job handler is correct, if we remove a taskName from our default configuration for validTaskNames this test will fail; but there is nothing wrong with the code.

It would be better to create a global variable map[string]**bool** and populate it with values you control in this test. Like so:

```
var taskName1, taskName2 = "dataset-api", "zebedee" // This can be set to anything and added to the map below

var taskNames = map[string]bool{
	taskName1: true,
	taskName2: true,
}
```

The reason to use bool over int, is to be able to collapse check of isPresent value in function `ParseTaskName`

Once using bool values instead of int:

change from
```
func ParseTaskName(taskNameStr string, taskNameValues map[string]int) error {
	_, isPresent := taskNameValues[taskNameStr]
	if isPresent == false {
		return apierrors.ErrTaskInvalidName
	}

	return nil
}
```
to 
```
// ParseTaskName returns ErrTaskInvalidName if it cannot find a valid TaskName for the given string representation
func ParseTaskName(taskName string, taskNames map[string]bool) error {
	if isPresent := taskNames[taskName]; !isPresent {
		return apierrors.ErrTaskInvalidName
	}

	return nil
}
```

features/steps/jobs_steps.go
```diff
@@ -98,7 +98,14 @@ func runJobsFeatureService(f *JobsFeature, err error, ctx context.Context, cfg *
 
 	serviceList := service.NewServiceList(initFunctions)
 	testIdentityClient := clientsidentity.New(cfg.ZebedeeURL)
-	f.svc, err = service.Run(ctx, cfg, serviceList, "1", "", "", svcErrors, testIdentityClient)
+	validTaskNames := strings.Split(cfg.TaskNameValues, ",")
+
+	//create map of valid task name values
+	taskNameValues := make(map[string]int)
```
nshumoogum, line 104: Like the unit test, I would hard code the task names in map for component test

api/jobs_test.go
cookel2: I had misunderstood what an enum is in go - thanks for explaining. I've made those changes now.

api/tasks_test.go
```diff
@@ -0,0 +1,185 @@
...
+	return models.Task{
+		JobID:       jobID,
+		LastUpdated: lastUpdated,
+		Links: &models.TaskLinks{
+			Self: self,
+			Job:  job,
+		},
+		NumberOfDocuments: numberOfDocuments,
+		TaskName:          taskName,
+	}, err
```
ChrysmOre, line 184: Usually you would explicitly return nil here rather than err, helps with clarity and uniformity

api/tasks.go
```diff
@@ -1,20 +1,18 @@
...
 	"github.com/ONSdigital/log.go/v2/log"
 	"github.com/gorilla/mux"
+	"net/http"
 )
 
 var invalidBodyErrorMessage = "invalid request body"
 
 // CreateTaskHandler returns a function that generates a new TaskName resource containing default values in its fields.
-func (api *API) CreateTaskHandler(ctx context.Context) http.HandlerFunc {
+func (api *API) CreateTaskHandler(taskNames map[string]bool) http.HandlerFunc {
```
ChrysmOre, line 15: If you added taskNames into the api struct you wouldn't need to create a function-that-returns-a-function and pass it as an argument.. Would make this section of the code a little easier to read (and is generally a better pattern to use IMO)

cookel2: Good idea, I prefer that pattern too now you've pointed it out, thanks.

PR #41, opened 2021-11-12
Draft PR to discuss code changes - DO NOT MERGE

nshumoogum:
### What

See title

### How to review

This is only open for TL and developer to discuss changes as part of sign off

### Who can review

Dont review!


swagger.yaml
```diff
@@ -119,6 +119,29 @@ paths:
...
+    get:
+      tags:
+        - "Private User"
+      summary: Get a list of tasks for a given job.
+      description: "Get a list of tasks, for a given search reindex job, which meet the criteria specified by the four query parameters."
+      parameters:
+        - $ref: '#/parameters/id'
+        - $ref: "#/parameters/limit"
+        - $ref: "#/parameters/offset"
+        - $ref: "#/parameters/sort"
```
nshumoogum, line 131: I wouldn't add sort parameter to the list endpoints in this application yet because we are not allowing API users to set sort in the query to affect the order of results.

Also the sort field is not returned in the response body currently, so its worth removing it from the swagger spec

features/getting_tasks.feature
```diff
@@ -0,0 +1,157 @@
+Feature: Getting a list of tasks
+
+  Scenario: Three Tasks exist in the Data Store and a get request returns them successfully
+
+    Given I use a service auth token "validServiceAuthToken"
+    And zebedee recognises the service auth token as valid
+    And I have generated a job in the Job Store
+    And I call POST /jobs/{id}/tasks using the generated id
```
nshumoogum, line 8: I would not use other endpoints to setup the data - if they are broke then this test will fail, alluding to Get /tasks not working.

You will want to generate the tasks like you've done for the job on line 7

mongo/data_store.go
```diff
@@ -166,6 +166,66 @@ func (m *JobStore) GetTask(ctx context.Context, jobID string, taskName string) (
...
+			log.Error(ctx, "error closing iterator", err)
+		}
+	}()
+
+	tasks := make([]models.Task, numTasks)
+	if err := iter.All(&tasks); err != nil {
+		return results, err
+	}
+
+	paginator := pagination.NewPaginator(m.cfg.DefaultLimit, m.cfg.DefaultOffset, m.cfg.DefaultMaxLimit)
```
nshumoogum, line 212: I would have moved the pagination validation before requesting data from mongo, if validation fails because of bad offset, limit values then return before making requests to the database. This will be a performance gain (more efficient).

Once validation occurs then call into the database using this method, you will need to pass in the offset and limit values into this method.

Again use the mongo db querying language to retrieve data instead of iterating all documents and working it out below as this again increases memory usage and will be slower to respond.
The above can be done using [Skip](https://docs.mongodb.com/compass/current/query/skip/) (represents offset); [Limit](https://docs.mongodb.com/compass/current/query/limit/) (represents offset) and [OrderBy](https://docs.mongodb.com/manual/reference/operator/meta/orderby/) (represent sort).

mongo/data_store.go
```diff
@@ -166,6 +166,66 @@ func (m *JobStore) GetTask(ctx context.Context, jobID string, taskName string) (
...
+
+	_, err := m.findJob(s, jobID, models.Job{})
+	if err != nil {
+		if err == mgo.ErrNotFound {
+			return results, ErrJobNotFound
+		}
+		return results, err
+	}
+
+	numTasks, _ := s.DB(m.Database).C(m.TasksCollection).Find(bson.M{"job_id": jobID}).Count()
```
nshumoogum, line 189: We should still handle this error here, we wouldn't want the code to panic in case we lose the connection to mongo.

mongo/data_store.go
```diff
@@ -350,3 +411,15 @@ func modifyJobs(jobs []models.Job, offset int, limit int) []models.Job {
 	}
 	return modifiedJobs
 }
+
+// modifyTasks takes a slice, of all the tasks in the data store, determined by the offset and limit values
+func modifyTasks(tasks []models.Task, offset int, limit int) []models.Task {
```
nshumoogum, line 416: This wont be needed if we rely on mongo to give us the correct tasks back, see earlier comment

features/getting_tasks.feature
cookel2: I did try doing it like that but in fact line 7 is using the Get /jobs endpoint - it's just that it's doing it within the jobs_steps rather than in the feature file. The thing that prevented me from doing it that way was that the body (to pass into each step) can only be specified in the feature file (as far as I could see).

swagger.yaml
cookel2: good point - done

mongo/data_store.go
cookel2: makes sense - done

mongo/data_store.go
cookel2: Thanks for this suggestion - it's good to know about those mongo db querying language methods. I've made those changes now.

mongo/data_store.go
cookel2: removed now

nshumoogum:
Closing draft, changes made against develop

PR #42, opened 2021-11-12
Feature/improve list tasks

cookel2:
# What has changed

Some improvements have been made to the new endpoint, which allows a GET request to be made, to list all the tasks for a particular search reindex job currently held in the data store. The endpoint is GET /jobs/{id}/tasks

The improvements are:
1. The sort field has been removed from the swagger spec. This is because we are not allowing API users to set sort in the query to affect the order of results.

2. When the requested tasks are counted there is potential for an error to occur if the connection to mongo is lost. That error was not previously being handled but now it is.

3. The pagination validation now occurs before the tasks are retrieved from mongo. This means that if the validation fails (because of bad offset or limit values) then it will return, and not make any database requests, which is more efficient.

4. The mongo dB querying language is now being used to better effect. Instead of retrieving all the tasks for a particular job, and iterating through them, it now uses the offset and limit parameters passed in and only returns the data that has been requested. It does not use an iterator but instead returns all the required data straight away.

5. Due to the improved use of the mongo dB querying language there is no need to use the modifyTasks function. So that has been removed.

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

Then hit this POST endpoint to create a job. Take note of the job id that it generates:
http://localhost:25700/jobs

Then hit this POST URL (using the job id noted above) e.g for a job with id 1b9a6881-8bf3-4729-8c59-62901c4688b8: 

{
  "task_name": "dataset-api",
  "number_of_documents": 18
}

Hit the same POST URL again to create a second task. But this time change the task_name value to zebedee (in the request body).

Finally the new GET endpoint can now be tested. It should list the 2 tasks:

# Who can review

Anyone but me

### DIFF
--- a/mongo/data_store.go
+++ b/mongo/data_store.go
@@ -186,13 +186,24 @@ func (m *JobStore) GetTasks(ctx context.Context, offsetParam string, limitParam 
 		return results, err
 	}
 
-	numTasks, _ := s.DB(m.Database).C(m.TasksCollection).Find(bson.M{"job_id": jobID}).Count()
+	numTasks := 0
+	numTasks, err = s.DB(m.Database).C(m.TasksCollection).Find(bson.M{"job_id": jobID}).Count()
+	if err != nil {
+		log.Error(ctx, "error counting tasks for given job id", err, log.Data{"job_id": jobID})
+		return results, err
+	}
 	log.Info(ctx, "number of tasks found in tasks collection", log.Data{"numTasks": numTasks})
 
 	if numTasks == 0 {
 		log.Info(ctx, "there are no tasks in the data store - so the list is empty")
 		results.TaskList = make([]models.Task, 0)
 		return results, nil
+	}
+
+	paginator := pagination.NewPaginator(m.cfg.DefaultLimit, m.cfg.DefaultOffset, m.cfg.DefaultMaxLimit)
+	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam, numTasks)
+	if err != nil {
+		return results, err
 	}
 
 	// Get all the tasks from the tasks collection and order them by lastupdated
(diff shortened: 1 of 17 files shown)


mongo/data_store.go
```diff
@@ -195,26 +200,18 @@ func (m *JobStore) GetTasks(ctx context.Context, offsetParam string, limitParam
...
-			log.Error(ctx, "error closing iterator", err)
-		}
-	}()
-
-	tasks := make([]models.Task, numTasks)
-	if err := iter.All(&tasks); err != nil {
-		return results, err
-	}
-
 	paginator := pagination.NewPaginator(m.cfg.DefaultLimit, m.cfg.DefaultOffset, m.cfg.DefaultMaxLimit)
```
nshumoogum, line 203: The pagination and validation shouldn't take place here, but should be called in the handlers for list endpoints.

Validation should always take place ahead of calling databases.

nshumoogum (changes requested):

mongo/data_store.go
cookel2: That's ok but the reason why I put it here is because the ValidatePaginationParameters needs to use the number of tasks, which can only be found out by calling the database beforehand.

nshumoogum: I disagree here, I don't believe we need the number of tasks to validate the request. If the number of tasks is less than the offset and limit value then we would return status 200 with 0 items in response body.

cookel2: Ok, just to be clear then, if the user entered an offset of 5 but there were only 3 tasks then we should just return a 200 status and an empty list?

cookel2: I've made the changes Nathan - please have another look when you have time. Thanks.

nshumoogum (approved):

PR #44, opened 2021-11-23
Feature/amend get jobs pagination

cookel2:
# What has changed

Changes have been made to the GET /jobs endpoint, which allows a GET request to be made, to list all the search reindex jobs currently held in the data store.

The changes are:

1. When the requested jobs are counted there is potential for an error to occur if the connection to mongo is lost. That error was not previously being handled but now it is.

2. If an offset parameter is passed in, which has a value greater than the total count of jobs in the database, instead of returning an error it will now return an empty list and a status of 200. This is to prevent unnecessary failures.

3. The pagination validation now occurs before the jobs are retrieved from mongo. This means that if the validation fails (because of bad offset or limit values) then it will return, and not make any database requests, which is more efficient.

4. The mongo dB querying language is now being used to better effect. Instead of retrieving all the jobs in the database, and iterating through them, it now uses the offset and limit parameters passed in and only returns the data that has been requested. It does not use an iterator but instead returns all the required data straight away.

5. Due to the improved use of the mongo dB querying language there is no need to use the modifyTasks function. So that has been removed.

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

Then hit the POST endpoint a few times to create some jobs:
http://localhost:25700/jobs

The changes can then be tested locally using this GET URL or similar: 
http://localhost:25700/jobs?offset=3&limit=100

# Who can review

Anyone but me

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -113,9 +113,9 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	offsetParam := req.URL.Query().Get("offset")
 	limitParam := req.URL.Query().Get("limit")
 
-	offset, limit, err := api.setUpPagination(w, offsetParam, limitParam, ctx)
+	offset, limit, err := api.setUpPagination(offsetParam, limitParam)
 	if err != nil {
-		log.Error(ctx, "setting up pagination failed", err)
+		log.Error(ctx, "pagination validation failed", err)
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
@@ -144,10 +144,9 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
-func (api *API) setUpPagination(w http.ResponseWriter, offsetParam string, limitParam string, ctx context.Context) (int, int, error) {
+func (api *API) setUpPagination(offsetParam string, limitParam string) (int, int, error) {
 	paginator := pagination.NewPaginator(api.cfg.DefaultLimit, api.cfg.DefaultOffset, api.cfg.DefaultMaxLimit)
-	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam)
-	return offset, limit, err
+	return paginator.ValidatePaginationParameters(offsetParam, limitParam)
 }
 
 // unlockJob unlocks the provided job lockID and logs any error with WARN state
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -117,10 +117,18 @@ func (api *API) GetJobHandler(ctx context.Context) http.HandlerFunc {
 func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
 	log.Info(ctx, "Entering handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.")
-	offsetParameter := req.URL.Query().Get("offset")
-	limitParameter := req.URL.Query().Get("limit")
+	offsetParam := req.URL.Query().Get("offset")
+	limitParam := req.URL.Query().Get("limit")
 
-	jobs, err := api.dataStore.GetJobs(ctx, offsetParameter, limitParameter)
+	paginator := pagination.NewPaginator(api.cfg.DefaultLimit, api.cfg.DefaultOffset, api.cfg.DefaultMaxLimit)
+	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam)
+	if err != nil {
+		log.Error(ctx, "pagination validation failed", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	jobs, err := api.dataStore.GetJobs(ctx, offset, limit)
 	if err != nil {
 		switch {
 		case badRequest[err]:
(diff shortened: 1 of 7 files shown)


api/jobs.go
```diff
@@ -151,6 +156,17 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (api *API) setUpPagination(w http.ResponseWriter, offsetParam string, limitParam string, ctx context.Context) (int, int, bool) {
```
rafahop, line 159: Can I suggest returning the error and deal with it in `GetJobsHandler`?

```suggestion
func (api *API) setUpPagination(w http.ResponseWriter, offsetParam string, limitParam string, ctx context.Context) (int, int, error) {
```

api/jobs.go
```diff
@@ -117,10 +117,15 @@ func (api *API) GetJobHandler(ctx context.Context) http.HandlerFunc {
...
-	jobs, err := api.dataStore.GetJobs(ctx, offsetParameter, limitParameter)
+	offset, limit, setUpPaginationFailed := api.setUpPagination(w, offsetParam, limitParam, ctx)
+	if setUpPaginationFailed {
+		return
+	}
+
+	jobs, err := api.dataStore.GetJobs(ctx, offset, limit)
 	if err != nil {
 		switch {
 		case badRequest[err]:
```
rafahop, line 131: I believe this case statement is pointless now and could go (together with the `badRequest` map)

rafahop (commented):
Nice one. It looks good to me but added a couple of comments.

I would also suggest adding test cases that use the `offset` and `limit` parameters? It would be particularly helpful having cases where those values are invalid and a 400 is returned

api/jobs.go
cookel2: yes that makes sense - done

api/jobs.go
cookel2: yes - deleted both

api/jobs.go
```diff
@@ -151,6 +143,16 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
...
 }
 
+func (api *API) setUpPagination(w http.ResponseWriter, offsetParam string, limitParam string, ctx context.Context) (int, int, error) {
+	paginator := pagination.NewPaginator(api.cfg.DefaultLimit, api.cfg.DefaultOffset, api.cfg.DefaultMaxLimit)
+	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam)
+	if err != nil {
+		log.Error(ctx, "pagination validation failed", err)
+		return 0, 0, err
+	}
+	return offset, limit, err
```
rafahop, line 153: Maybe consider letting the client log the error however they want? 
It would simplify the code a bit and seems to be consistent with how we deal with these errors (`log.Error` followed by `http.Error`). 
We wouldn't need to pass the context as a parameter and simply do: 
```suggestion
	return paginator.ValidatePaginationParameters(offsetParam, limitParam)
```

cookel2: yes, I was unsure where to put the logging but I like your suggestion of simplifying the code - done

api/jobs.go
```diff
@@ -151,6 +144,12 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	}
 }
 
+func (api *API) setUpPagination(w http.ResponseWriter, offsetParam string, limitParam string, ctx context.Context) (int, int, error) {
+	paginator := pagination.NewPaginator(api.cfg.DefaultLimit, api.cfg.DefaultOffset, api.cfg.DefaultMaxLimit)
+	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam)
+	return offset, limit, err
```
rafahop, line 150: No need to store the variables?

```suggestion
	return paginator.ValidatePaginationParameters(offsetParam, limitParam)
```

api/tasks.go
```diff
@@ -108,10 +107,8 @@ func (api *API) GetTasksHandler(w http.ResponseWriter, req *http.Request) {
 	id := vars["id"]
 	logData := log.Data{"job_id": id}
 
-	paginator := pagination.NewPaginator(api.cfg.DefaultLimit, api.cfg.DefaultOffset, api.cfg.DefaultMaxLimit)
-	offset, limit, err := paginator.ValidatePaginationParameters(offsetParam, limitParam)
+	offset, limit, err := api.setUpPagination(w, offsetParam, limitParam, ctx)
 	if err != nil {
-		log.Error(ctx, "pagination validation failed", err)
```
rafahop, line 114: I think we should bring this line back now that we are not logging the error within `setUpPagination`

api/jobs_test.go
```diff
@@ -311,14 +323,109 @@ func TestGetJobsHandler(t *testing.T) {
...
+
+			Convey("Then a bad request error is returned with status code 400", func() {
+				So(resp.Code, ShouldEqual, http.StatusBadRequest)
+				errMsg := strings.TrimSpace(resp.Body.String())
+				So(errMsg, ShouldEqual, expectedOffsetErrorMsg)
+			})
+		})
+
+		Convey("When a request is made to get a list of jobs with a limit that is negative", func() {
+			req := httptest.NewRequest("GET", "http://localhost:25700/jobs?offset=0&limit=-1", nil)
```
rafahop, line 397: Thanks for adding these tests 🙂  
For completeness, I would probably add cases for negative value in `offset` and non-numeric value in `limit`?

api/jobs.go
cookel2: no I guess not, good point - removed

api/tasks.go
cookel2: yes, that makes sense - done

api/jobs_test.go
cookel2: no problem - it was quick to do

rafahop (approved):

PR #48, opened 2021-12-17
Feature/extend post job endpoint

cookel2:
# What has changed

The POST /jobs endpoint functionality has been extended; rather than just creating a default Job Resource, in the Mongo data store, and returning its JSON representation, it now does as follows:
- creates a Job Resource in the Mongo data store
- calls the Search API POST /search endpoint to create a new Elasticsearch index using ES version 7.10 (the version for site wide search)
- extracts the name of the new search index from the Search API response. NB. The name will be of the form ons{timestamp} where the timestamp is the number of microseconds since epoch time
- updates the search_index_name, of the Job Resource, to be the name of the new search index
- if an error occurs, which prevents a new Elasticsearch index from being created and/or prevents its name from being extracted, then it updates the status, of the Job Resource, to "failed"
- returns a JSON representation of the Job Resource

Also, the following two new component tests have been added to posting_a_job.feature:
- Scenario: The connection to search API is lost and a post request returns a job state of failed
- Scenario: The search API is not pointing to the correct version of ES and a post request returns a job state of failed

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

 make debug

http://localhost:25700/jobs

Check the returned JSON and also look in Mongo DB to make sure that the name begins with "ons" and the status is "created". Then try stopping the Search API, or changing its ELASTIC_SEARCH_URL to the old version (http://localhost:9200), then hit the POST /jobs endpoint again and make sure that the name is "Default Search Index Name" and the status is "failed". 

# Who can review

Anyone but me


Marianasilvago (approved):
LGTM

PR #49, opened 2021-12-22
Update audit docker image version to be consistent

rahulPalliyalil:
### What

Just an update to the audit docker image version to be consistent with the other layers

### How to review

n/a

### Who can review

Anyone except me

### DIFF
--- a/ci/audit.yml
+++ b/ci/audit.yml
@@ -5,7 +5,7 @@ image_resource:
   type: docker-image
   source:
     repository: onsdigital/dp-concourse-tools-nancy
-    tag: 1.17.4
+    tag: latest
 
 inputs:
   - name: dp-search-reindex-api
--- a/ci/audit.yml
+++ b/ci/audit.yml
@@ -5,7 +5,7 @@ image_resource:
   type: docker-image
   source:
     repository: onsdigital/dp-concourse-tools-nancy
-    tag: latest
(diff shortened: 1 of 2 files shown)


ci/audit.yml
```diff
@@ -5,7 +5,7 @@ image_resource:
   type: docker-image
   source:
     repository: onsdigital/dp-concourse-tools-nancy
-    tag: latest
+    tag: 1.17.4
```
nshumoogum, line 8: So for both audit and lint yml files we use latest, see https://github.com/ONSdigital/dp-hello-world/tree/master/dp-hello-world-controller/ci

I believe the reasoning is to capture issues - I think what we really want is both the go version we are using and the latest version assuming it is the same major version of Go (is backward compatible)

nshumoogum (changes requested):

ci/audit.yml
rahulPalliyalil: In my opinion, atleast for lint we have to fix the version . The reason behind is that if we use the latest tag, then the if the go version changes to say 18 (which is in beta right now), the lint will flag issues that was passing in previous builds.For example if a package is depreciated, so to avoid that i think it is best to stick with the version for the linter atleast . I can update to the latest as per the comment but let me know your thoughts .

nshumoogum (approved):

PR #50, opened 2022-01-04
Feature/fix service auth token

cookel2:
# What has changed

The current standard (for the service auth token) is not to set a default value for this configuration value and instead developers export their own value and create the equivalent token in zebedee content.

Therefore the default value for the SERVICE_AUTH_TOKEN has been removed from the config file. 

Also, the README has been updated to include details of new dependencies and environment variables (mostly required for calling the Search API from the POST /jobs endpoint).

# How to review

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

 make debug

http://localhost:25700/jobs

Check the returned JSON and also look in Mongo DB to make sure that the name begins with "ons" and the status is "created". 

# Who can review

Anyone but me

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,30 @@ Provides detail about search reindex jobs and enables creation and running of th
 
 * Set up dependencies locally as follows:
 
-In dp-compose run `docker-compose up -d` to run MongoDB on port 27017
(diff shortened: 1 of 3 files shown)


rahulPalliyalil (approved):

Marianasilvago (commented):
LGTM

PR #51, opened 2022-01-06
Release/0.14.0

cookel2:
# What has changed

These are the changes in this release:

The following amendments have been made to the POST /jobs endpoint, which allows a POST request to be made, to create a new Job Resource in the Mongo data store:

1. The POST /jobs handler now calls the Search API, after creating the new Job Resource, to trigger the creation of a new Elastic Search index.

2. If the new Elastic Search index is created successfully then the new Job Resource, and the JSON to be returned by the POST /jobs endpoint, are both updated with the new index name (that was returned by the Search API). However, if the new Elastic Search index is not created successfully, then those things do not have the new index name updated but instead have the job state updated to 'failed'.

3. The healthchecks now include a check for the Search API.

4. The version of dp-component-test being used is now v0.6.3

5. golangci linter capabilities have been added and set as a makefile target

6. The version of golang being used is now 1.17.4

# How to review

Check only expected commits present

NB. The new functionality can be tested locally as follows:

 docker-compose up -d

 vault server -dev

 ./run.sh

In the dp-search-api repo set the ELASTIC_SEARCH_URL environment variable as follows (to use the Site Wide ElasticSearch):

Also in the dp-search-api repo run make debug

Make sure that you have a valid local SERVICE_AUTH_TOKEN environment variable value; if not then set one up by following these instructions: https://github.com/ONSdigital/zebedee

 make debug

Then hit the POST /jobs endpoint and see if the search_index_name gets updated to something like this: ons1638363874110115

http://localhost:25700/jobs

# Who can review

Anyone but me


api/api_test.go
```diff
@@ -23,13 +24,14 @@ func TestSetup(t *testing.T) {
...
-		api := api.Setup(context.Background(), mux.NewRouter(), &mock.DataStorerMock{}, &mock.AuthHandlerMock{}, taskNames, cfg)
+		apiClient := api.Setup(mux.NewRouter(), &mock.DataStorerMock{}, &mock.AuthHandlerMock{}, taskNames, cfg, httpClient, &mock.IndexerMock{})
 
 		Convey("When created the following routes should have been added", func() {
-			So(hasRoute(api.Router, "/jobs", "POST"), ShouldBeTrue)
-			So(hasRoute(api.Router, "/jobs/{id}", "GET"), ShouldBeTrue)
-			So(hasRoute(api.Router, "/jobs", "GET"), ShouldBeTrue)
+			So(hasRoute(apiClient.Router, "/jobs", "POST"), ShouldBeTrue)
+			So(hasRoute(apiClient.Router, "/jobs/{id}", "GET"), ShouldBeTrue)
+			So(hasRoute(apiClient.Router, "/jobs", "GET"), ShouldBeTrue)
```
goofballLogic, line 34: just out of interest - why aren't we testing for the other setup routes here?

Makefile
```diff
@@ -24,7 +24,8 @@ debug:
 
 .PHONY: lint
 lint:
-	exit
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.43.0
+	golangci-lint run ./...
```
goofballLogic, line 28: Didn't work locally (needed to brew install instead) but as CI is working this seems ok

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
 	serverErrorMessage = "internal server error"
 )
 
-// CreateJobHandler returns a function that generates a new Job resource containing default values in its fields.
-func (api *API) CreateJobHandler(ctx context.Context) http.HandlerFunc {
-	return func(w http.ResponseWriter, req *http.Request) {
-		ctx := req.Context()
+// CreateJobHandler generates a new Job resource and a new ElasticSearch index associated with it	.
+func (api *API) CreateJobHandler(w http.ResponseWriter, req *http.Request) {
```
goofballLogic, line 33: Correct me if I'm wrong, but this is no longer "creating" the job handler - it is the job handler, no?

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
...
 
-		newJob, err := api.dataStore.CreateJob(ctx, id)
-		if err != nil {
-			log.Error(ctx, "creating and storing job failed", err)
-			if err == mongo.ErrExistingJobInProgress {
-				http.Error(w, "existing reindex job in progress", http.StatusConflict)
-			} else {
+	log.Info(ctx, "creating new index in ElasticSearch via the Search API")
+	serviceAuthToken := "Bearer " + api.cfg.ServiceAuthToken
+	searchAPISearchURL := api.cfg.SearchAPIURL + "/search"
```
goofballLogic, line 51: do we have a rule/utility protecting against the possibility that SearchAPIURL already terminates with a `/` (or do we just ignore double slashes?)

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
...
+	reindexResponse, err := api.reindex.CreateIndex(ctx, serviceAuthToken, searchAPISearchURL, api.httpClient)
+	if err != nil {
+		log.Error(ctx, "error occurred when connecting to Search API", err)
+		if newJob != (models.Job{}) {
```
goofballLogic, line 55: the `http.Error` and `return` are inside this if block so if it doesn't meet this condition, won't it just fall through to successful response?

api/jobs.go
```diff
@@ -15,6 +16,10 @@ import (
 	uuid "github.com/satori/go.uuid"
 )
 
+const (
+	jobStateFailed = "failed"
```
goofballLogic, line 20: would this be better placed in the model - e.g. in a `Failed` method?

goofballLogic: or in a package like `apierrors` ?

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
...
 		}
-
-		w.Header().Set("Content-Type", "application/json")
-		jsonResponse, err := json.Marshal(newJob)
-		if err != nil {
-			log.Error(ctx, "marshalling response failed", err)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
+	} else if reindexResponse.StatusCode != 201 {
+		log.Error(ctx, "error occurred in post search http request", err)
```
goofballLogic, line 66: Here `err` will be nil i think?

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
...
-			return
 		}
-
-		w.Header().Set("Content-Type", "application/json")
-		jsonResponse, err := json.Marshal(newJob)
-		if err != nil {
-			log.Error(ctx, "marshalling response failed", err)
-			http.Error(w, serverErrorMessage, http.StatusInternalServerError)
-			return
+	} else if reindexResponse.StatusCode != 201 {
```
goofballLogic, line 65: what about other success codes? e.g. can it return a 200?

api/jobs.go
goofballLogic, line 55: this block of code is repeated three times (i think?) can we DRY it out?

api/jobs.go
```diff
@@ -24,84 +29,125 @@ var (
...
+				log.Error(ctx, "setting state to failed has failed", setStateErr)
+				http.Error(w, serverErrorMessage, http.StatusInternalServerError)
+				return
+			}
 		}
-
-		w.WriteHeader(http.StatusCreated)
-		_, err = w.Write(jsonResponse)
+	} else {
+		newJob, err = api.updateSearchIndexName(ctx, reindexResponse, err, newJob, id)
```
goofballLogic, line 78: `err` passed in here will always be nil won't it? Can we just pass `nil` to clarify this?

api/jobs.go
```diff
@@ -158,52 +204,86 @@ func (api *API) unlockJob(ctx context.Context, lockID string) {
...
+			log.Error(ctx, "error closing http response body", err)
 		}
+	}
+}
 
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
+// updateSearchIndexName calls the GetIndexNameFromResponse function, in the reindex package, to get the index name that was returned by the Search API.
+// It then calls the UpdateIndexName function, in the mongo package, to update the search_index_name value in the relevant Job Resource in the data store.
+func (api *API) updateSearchIndexName(ctx context.Context, reindexResponse *http.Response, searchErr error, newJob models.Job, id string) (models.Job, error) {
```
goofballLogic, line 266: `searchErr` appears unused?

api/jobs.go
```diff
@@ -158,52 +204,86 @@ func (api *API) unlockJob(ctx context.Context, lockID string) {
...
+	defer closeResponseBody(ctx, reindexResponse)
```
goofballLogic, line 267: is it better practice to defer closeResponseBody at the point where we know it was created (before calling this function?)

api/jobs_test.go
```diff
@@ -57,27 +60,40 @@ func TestCreateJobHandler(t *testing.T) {
...
+		httpClient := dpHTTP.NewClient()
+
+		apiInstance := api.Setup(mux.NewRouter(), dataStorerMock, &apiMock.AuthHandlerMock{}, taskNames, cfg, httpClient, indexerMock)
 
 		Convey("When a new reindex job is created and stored", func() {
 			req := httptest.NewRequest("POST", "http://localhost:25700/jobs", nil)
 			resp := httptest.NewRecorder()
 
-			createJobHandler.ServeHTTP(resp, req)
+			apiInstance.CreateJobHandler(resp, req)
```
goofballLogic, line 87: elsewhere we appear to be calling through `ServeHTTP` - how come this one is different?

api/jobs_test.go
```diff
@@ -57,27 +60,40 @@ func TestCreateJobHandler(t *testing.T) {
...
 			Convey("Then the newly created search reindex job is returned with status code 201", func() {
 				So(resp.Code, ShouldEqual, http.StatusCreated)
-				payload, err := ioutil.ReadAll(resp.Body)
+				payload, err := io.ReadAll(resp.Body)
 				So(err, ShouldBeNil)
 				newJob := models.Job{}
 				err = json.Unmarshal(payload, &newJob)
 				So(err, ShouldBeNil)
-				expectedJob, err := models.NewJob(validJobID1)
+				expectedJob, err := ExpectedJob(validJobID1, zeroTime, 0, zeroTime, zeroTime, zeroTime, "ons1638363874110115", "created", 0, 0)
```
goofballLogic, line 96: I'm not an expert on Go conventions but this function with loads of arguments would be undesirable in other languages. Can't we just do `expectedJob, err := models.NewJob(validJobID1)` as done elsewhere? If the test is that we want to check all fields are return the same then we would need to assign them to something other than the defaults wouldn't we?

features/getting_jobs.feature
```diff
@@ -2,22 +2,23 @@ Feature: Getting a list of jobs
...
+      | last_updated      | Not in the future                      |
+      | links: tasks      | http://{bind_address}/jobs/{id}/tasks  |
+      | links: self       | http://{bind_address}/jobs/{id}        |
+      | search_index_name | ons{date_stamp}                        |
     And each job should also contain the following values:
       | number_of_tasks                 | 0                         |
       | reindex_completed               | 0001-01-01T00:00:00Z      |
       | reindex_failed                  | 0001-01-01T00:00:00Z      |
       | reindex_started                 | 0001-01-01T00:00:00Z      |
-      | search_index_name               | Default Search Index Name |
```
goofballLogic, line 20: why did we stop testing for this value?

features/posting_a_job.feature
```diff
@@ -2,28 +2,48 @@ Feature: Posting a job
...
     Then the HTTP status code should be "500"
+
+  Scenario: The connection to search API is lost and a post request returns a job state of failed
+
+    Given the search reindex api loses its connection to the search api
+    When I POST "/jobs"
+    """
+    """
+    Then the response should contain a state of "failed"
+    And the HTTP status code should be "201"
```
goofballLogic, line 40: why would we want a status code of 201 if the POST fails?

features/posting_a_job.feature
```diff
@@ -2,28 +2,48 @@ Feature: Posting a job
...
+
+  Scenario: The search API is not pointing to the correct version of ES and a post request returns a job state of failed
+
+    Given the search reindex api is not working correctly because it is pointing to the old version of ES
+    When I POST "/jobs"
+    """
+    """
+    Then the response should contain a state of "failed"
+    And the HTTP status code should be "201"
```
goofballLogic, line 49: ditto

mongo/data_store.go
```diff
@@ -33,6 +33,36 @@ type JobStore struct {
 	cfg             *config.Config
 }
 
+// Init creates a new mgo.Session with a strong consistency and a write mode of "majority".
+func (m *JobStore) Init(ctx context.Context, cfg *config.Config) (err error) {
```
goofballLogic, line 37: are we expected to do some sort of testing for this?

reindex/reindex.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// CreateIndex calls the Search API via the Do function of the dp-net/http/Clienter. It passes in the ServiceAuthToken to identify itself, as the Search Reindex API, to the Search API.
+func (r *Reindex) CreateIndex(ctx context.Context, serviceAuthToken, searchAPISearchURL string, httpClient dphttp.Clienter) (*http.Response, error) {
+	req, err := http.NewRequest(http.MethodPost, searchAPISearchURL, http.NoBody)
+	if err != nil {
+		return nil, errors.New("failed to create the request for post search")
+	}
+	if err := headers.SetServiceAuthToken(req, serviceAuthToken); err != nil {
+		//TODO: ideally this needs to return an error when it couldn't set a service auth token.
```
goofballLogic, line 31: yeah this seems important

reindex/reindex.go
```diff
@@ -0,0 +1,61 @@
...
+	readBodyFailedMsg := "failed to read response body"
+	unmarshalBodyFailedMsg := "failed to unmarshal response body"
+
+	if err != nil {
+		log.Error(ctx, readBodyFailedMsg, err, logData)
+		return "", errors.New(readBodyFailedMsg)
+	}
+
+	if len(b) == 0 {
+		b = []byte("[response body empty]")
```
goofballLogic, line 50: shouldn't we just return an error here?

service/service_test.go
```diff
@@ -332,14 +329,14 @@ func TestClose(t *testing.T) {
...
 
 			err = svc.Close(context.Background())
 			So(err, ShouldNotBeNil)
 			So(hcMock.StopCalls(), ShouldHaveLength, 1)
 			So(failingServerMock.ShutdownCalls(), ShouldHaveLength, 1)
-			So(mongoDbMock.CloseCalls(), ShouldHaveLength, 1)
+			So(mongoDBMock.CloseCalls(), ShouldHaveLength, 1)
 		})
 	})
 }
```
goofballLogic, line 342: don't we need some tests for the new search service client health check?

reindex/reindex.go
```diff
@@ -0,0 +1,61 @@
+package reindex
```
goofballLogic, line 1: are there tests for this somewhere

api/jobs.go
goofballLogic: maybe for next time

api/jobs.go
goofballLogic: maybe next time

api/jobs.go
goofballLogic: maybe next time (would require dataStore to return nil BUT ALSO not return an error)

goofballLogic: maybe next time

api/jobs.go
goofballLogic: maybe next time

api/jobs.go
goofballLogic: maybe next time we will log status code instead of (nil err)

api/jobs.go
goofballLogic: maybe next time

api/jobs.go
goofballLogic: maybe next time

api/jobs.go
goofballLogic: maybe next time. If it's needed there is another logic branch where this isn't happening.

api/jobs_test.go
goofballLogic: for next time - use apiInstance.Router.ServeHTTP instead to cover routing etc.

api/jobs_test.go
goofballLogic: Somebody else thinks this is the right way (?)

features/posting_a_job.feature
goofballLogic: creation of a failed job is success (!)

goofballLogic (approved):
all comments can wait until next time

api/jobs.go
cookel2: I've amended this now

api/jobs.go
cookel2: I've moved the constant to the model and renamed it to JobStateFailed so that it gets exported. I've also added a JobStateCreated as that is the other possible job state that the service currently uses.

api/jobs.go
cookel2: done

api/jobs.go
cookel2: I've refactored due to earlier comment

api/jobs.go
cookel2: now removed from function so no need to pass in

api/jobs.go
cookel2: done

reindex/reindex.go
cookel2: done

mongo/data_store.go
cookel2: not as far as I know

reindex/reindex.go
cookel2: agreed, done

service/service_test.go
cookel2: I think these are now covered?

reindex/reindex.go
cookel2: yes there are component tests for this, which test the POST /jobs endpoint

api/jobs.go
goofballLogic: looks good

api/jobs.go
goofballLogic: seems to also be a reference to literal "failed" on line 285

goofballLogic: oh and 287 maybe?

api/jobs.go
goofballLogic: looks good

api/jobs.go
goofballLogic: i think this one can still be done.

api/jobs.go
goofballLogic: looks good

api/jobs.go
goofballLogic: looks good

api/jobs.go
goofballLogic: ok no big deal

reindex/reindex.go
goofballLogic: looks good

reindex/reindex.go
goofballLogic: looks good

service/service_test.go
goofballLogic: more or less by producer_test I think

reindex/reindex.go
goofballLogic: ok

api/jobs.go
goofballLogic: looks good

PR #55, opened 2022-01-24
Produce reindex requested event

cookel2:
# What has changed

The POST /jobs endpoint functionality has now been extended to send a kafka event to the reindex-requested topic. So it now does as follows:
- creates a Job Resource in the Mongo data store
- calls the Search API POST /search endpoint to create a new Elasticsearch index using ES version 7.10 (the version for site wide search)
- sends a kafka event to the reindex-requested topic
- returns a JSON representation of the Job Resource

NB. The reindex-requested kafka event has the following format:
"event": {
    "JobID": "8a884a54-2400-4867-b769-26fee4377da0",
    "SearchIndex": "ons1643023710759669",
    "TraceID": "IYMhveeetjOiBxds"
},

Also, in the posting_a_job.feature, the following scenario has been extended to check that the kafka event is sent to the producer's output channel:
- Scenario: Job is posted successfully

And unit tests have been added to test the following failure scenarios:
- produce reindex request is called but the marshaller returns an error
- produce reindex request is called but it fails to send the marshalled event to the producer's blocked Output channel

# How to review

In dp-compose run MongoDB, ZooKeeper, and Kafka, as follows:

 docker-compose up -d

 vault server -dev

 ./run.sh

 make debug

 make debug

http://localhost:25700/jobs

Check the logs in the console and make sure that you see something like this:

{
  "created_at": "2022-01-24T12:35:37.023027Z",
  "data": {
    "reindexRequestedEvent": {
      "JobID": "5b77cfb7-a596-4e19-ba30-f3109ad2e6f2",
      "SearchIndex": "ons1643027736707602",
      "TraceID": "fUkbTvaRtCqMGLIE"
    }
  },
  "event": "sending reindex-requested event",
  "namespace": "dp-search-reindex-api",
  "severity": 3
}

# Who can review

Anyone but me


main_test.go
```diff
@@ -76,6 +78,14 @@ func (f *ComponentTest) InitializeTestSuite(ctx *godog.TestSuiteContext) {
 }
 func TestComponent(t *testing.T) {
 	if *componentFlag {
+		log.SetDestination(io.Discard, io.Discard)
+		golog.SetOutput(io.Discard)
+		// Commented out pending fix for log.go library:
```
goofballLogic, line 83: I think this is now fixed (https://github.com/ONSdigital/log.go/pull/34) but would need checking...

schema/schema.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"github.com/ONSdigital/dp-kafka/v2/avro"
+)
+
+var reindexRequested = `{
+	"type": "record",
+	"name": "reindex-requested",
+	"fields": [
+		{"name": "job_id", "type": "string", "default": ""},
+    	{"name": "search_index", "type": "string", "default": ""},
```
goofballLogic, line 12: indentation for lines 12 and 13 is a mix of tabs and spaces, so they're not aligned to line 11 which is two tabs

goofballLogic (approved):

goofballLogic (approved):

PR #67, opened 2022-03-17
sdk client implementation

bpathak-ons:
### What

* SDK client initial implementation

### How to review

* code sense check
* CI pass

### Who can review

* anyone


clients/v1/client.go
```diff
@@ -0,0 +1,100 @@
...
+	hcClient := healthcheck.NewClientWithClienter(service, searchReindexURL, clienter)
+
+	return &Client{
+		hcClient,
+	}
+}
+
+// NewWithHealthClient creates a new instance of Client,
+// reusing the URL and Clienter from the provided health check client.
+func NewWithHealthClient(hcCli *healthcheck.Client) *Client {
```
nshumoogum, line 49: Rename method to `NewClientWithHealthcheck`?

clients/v1/client.go
```diff
@@ -0,0 +1,100 @@
...
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/ONSdigital/log.go/v2/log"
+
+	healthcheck "github.com/ONSdigital/dp-api-clients-go/v2/health"
+	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
+	dphttp "github.com/ONSdigital/dp-net/http"
+	dprequest "github.com/ONSdigital/dp-net/request"
```
nshumoogum, line 14: And again dp-net/v2/request

clients/v1/client.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+// Checker calls search-reindex health endpoint and returns a check object to the caller.
+func (c *Client) Checker(ctx context.Context, check *health.CheckState) error {
+	return c.hcCli.Checker(ctx, check)
+}
+
+// Get returns a response for the requested uri in searchreindexapi
+func (c *Client) Get(ctx context.Context, userAccessToken, path string) ([]byte, error) {
```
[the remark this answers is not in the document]
nshumoogum: For this PR I would remove the Get methods, so `Get`, `get` and `GetWithHeaders`

sdk/client_test.go
```diff
@@ -0,0 +1,67 @@
...
+		req := &http.Request{}
+		headers := &Headers{
+			ETag: "dsalfhjsadf",
+		}
+
+		Convey("when calling the Add method on Headers", func() {
+			headers.Add(req)
+
+			Convey("then an ETag header is set on the request", func() {
+				So(req.Header, ShouldBeEmpty)
```
cookel2, line 28: I think it would be good to add a TODO here e.g.
```suggestion
                                 //TODO Replace the 'ShouldBeEmpty' with the expected header when the ETag header has been implemented
				So(req.Header, ShouldBeEmpty)
```

sdk/client_test.go
```diff
@@ -0,0 +1,67 @@
...
+		req := &http.Request{}
+		headers := &Headers{
+			IfMatch: "*",
+		}
+
+		Convey("when calling the Add method on Headers", func() {
+			headers.Add(req)
+
+			Convey("then an If-Match header is set on the request", func() {
+				So(req.Header, ShouldBeEmpty)
```
cookel2, line 43: Similarly I think it would be good to add a TODO here. Otherwise this test is passing, even though it should be failing as adding the If-Match header has not been implemented yet, and there's nothing explaining why.

sdk/client_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	authHeader = "Authorization"
+	authPrefix = "Bearer "
+)
+
+func TestHeaders_Add(t *testing.T) {
+	t.Parallel()
+
+	Convey("given the sdk Headers struct contains a value for ETag", t, func() {
```
cookel2, line 18: Then Given, When, and Then clauses are usually initcapped. I thought that was a standard?

sdk/v1/client.go
```diff
@@ -0,0 +1,156 @@
...
+			Err:  fmt.Errorf("failed to call search reindex api, error is: %v", err),
+			Code: http.StatusInternalServerError,
+		}
+	}
+	defer func() {
+		err = closeResponseBody(ctx, resp)
+	}()
+
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= 400 {
+		fmt.Printf("did we get here: %v", resp.StatusCode)
```
cookel2, line 126: Please delete line 126.

cookel2 (changes requested):
just a few comments, thanks

sdk/client_test.go
bpathak-ons: nope, the test will fail once eEag is implemented

sdk/client_test.go
bpathak-ons: nope, same as above, the test will fail once implemented

cookel2: I realise that the test will fail once implemented. Similarly to above, I was trying to suggest that you add a TODO comment to explain that this will need to be updated once the If-Match is implemented. But I don't mind if you leave it as is.

cookel2 (approved):

nshumoogum (approved):

PR #70, opened 2022-03-29
client implementation for task counts

bpathak-ons:
### What

* client for endpoint /jobs/{id}/tasks
* fix failing junit for timestamp error
* fix for lint CI error due to golang version

### How to review

* code sense check
* CI pass

### Who can review

* anyone

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -29,6 +29,12 @@ type Options struct {
 	Sort   string
 }
 
+// TaskNames is list of possible tasks associated with a job
+var TaskNames = map[string]string{
+	"zebedee":     "zebedee",
+	"dataset-api": "dataset-api",
+}
+
 func (h *Headers) Add(req *http.Request) {
 	if h == nil {
 		return
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -13,6 +13,7 @@ import (
 
 type Client interface {
 	PostJob(ctx context.Context, headers Headers) (models.Job, error)
+	PostTasksCount(ctx context.Context, headers Headers, jobID string) (models.Task, error)
 }
 
 type Headers struct {
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -17,7 +17,7 @@ type Client interface {
 	Checker(ctx context.Context, check *health.CheckState) error
 	Health() *healthcheck.Client
 	PostJob(ctx context.Context, headers Headers) (models.Job, error)
-	PostTasksCount(ctx context.Context, headers Headers, jobID string) (models.Task, error)
+	PostTasksCount(ctx context.Context, headers Headers, jobID string, payload []byte) (models.Task, error)
 	URL() string
 }
 
--- a/sdk/v1/client.go
+++ b/sdk/v1/client.go
@@ -81,6 +81,31 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (models.
 	return job, nil
 }
 
+// PostTasksCount Updates tasks count for processing
+func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string) (models.Task, error) {
+	var task models.Task
+
+	if headers.ServiceAuthToken == "" {
+		headers.ServiceAuthToken = cli.serviceToken
+	}
+
+	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
+
+	b, err := cli.callReindexAPI(ctx, path, http.MethodPost, headers, nil)
+	if err != nil {
+		return task, err
+	}
+
+	if err = json.Unmarshal(b, &task); err != nil {
+		return task, apiError.StatusError{
+			Err:  fmt.Errorf("failed to unmarshal bytes into reindex job, error is: %v", err),
+			Code: http.StatusInternalServerError,
+		}
+	}
+
+	return task, nil
+}
+
 func (cli *Client) callReindexAPI(ctx context.Context, path, method string, headers client.Headers, payload []byte) ([]byte, error) {
 	URL, err := url.Parse(path)
 	if err != nil {
--- a/sdk/v1/client.go
+++ b/sdk/v1/client.go
@@ -82,7 +82,7 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (models.
 }
 
 // PostTasksCount Updates tasks count for processing
-func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string) (models.Task, error) {
+func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string, payload []byte) (models.Task, error) {
 	var task models.Task
 
 	if headers.ServiceAuthToken == "" {
@@ -91,7 +91,7 @@ func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, j
 
 	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
 
-	b, err := cli.callReindexAPI(ctx, path, http.MethodPost, headers, nil)
+	b, err := cli.callReindexAPI(ctx, path, http.MethodPost, headers, payload)
 	if err != nil {
 		return task, err
 	}
--- a/sdk/v1/client_test.go
+++ b/sdk/v1/client_test.go
@@ -62,11 +62,11 @@ func TestClient_HealthChecker(t *testing.T) {
 		searchReindexClient := newSearchReindexClient(httpClient)
 		check := initialState
 
-		Convey("when search-reindexClient.Checker is called", func() {
+		Convey("When search-reindexClient.Checker is called", func() {
 			err := searchReindexClient.Checker(ctx, &check)
 			So(err, ShouldBeNil)
 
-			Convey("then the expected check is returned", func() {
+			Convey("Then the expected check is returned", func() {
 				So(check.Name(), ShouldEqual, service)
 				So(check.Status(), ShouldEqual, health.StatusCritical)
 				So(check.StatusCode(), ShouldEqual, 0)
@@ -76,7 +76,7 @@ func TestClient_HealthChecker(t *testing.T) {
 				So(*check.LastFailure(), ShouldHappenAfter, timePriorHealthCheck)
 			})
 
-			Convey("and client.Do should be called once with the expected parameters", func() {
+			Convey("And client.Do should be called once with the expected parameters", func() {
 				doCalls := httpClient.DoCalls()
 				So(doCalls, ShouldHaveLength, 1)
 				So(doCalls[0].Req.URL.Path, ShouldEqual, path)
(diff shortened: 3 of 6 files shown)


sdk/v1/client.go
```diff
@@ -81,6 +81,31 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (models.
...
+func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string) (models.Task, error) {
+	var task models.Task
+
+	if headers.ServiceAuthToken == "" {
+		headers.ServiceAuthToken = cli.serviceToken
+	}
+
+	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
+
+	b, err := cli.callReindexAPI(ctx, path, http.MethodPost, headers, nil)
```
cookel2, line 94: The body should not be nil. It should be the details of the task to create/update, which is defined in the models package by:
 type TaskToCreate struct {
	TaskName          string `json:"task_name"`
	NumberOfDocuments int    `json:"number_of_documents"`
}

sdk/v1/client_test.go
```diff
@@ -208,3 +208,95 @@ func TestClient_PostJob(t *testing.T) {
...
+			})
+
+			Convey("And client.Do should be called once with the expected parameters", func() {
+				doCalls := httpClient.DoCalls()
+				So(doCalls, ShouldHaveLength, 1)
+				So(doCalls[0].Req.URL.Path, ShouldEqual, pathToCheck)
+			})
+		})
+	})
+	Convey("given a 404 response", t, func() {
```
cookel2, line 281: Please can these gherkin key words also begin with an uppercase letter as that seems to be the standard practice.
```suggestion
	Convey("Given a 404 response", t, func() {
```

sdk/v1/client_test.go
```diff
@@ -208,3 +208,95 @@ func TestClient_PostJob(t *testing.T) {
...
+		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusConflict}, nil)
+		searchReindexClient := newSearchReindexClient(httpClient)
+
+		Convey("when search-reindexClient.PostJob is called", func() {
```
cookel2, line 285: This needs to say PostTasksCount instead of PostJob.
```suggestion
		Convey("When search-reindexClient.PostTasksCount is called", func() {
```

sdk/v1/client_test.go
```diff
@@ -208,3 +208,95 @@ func TestClient_PostJob(t *testing.T) {
...
+			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID)
+			So(err, ShouldNotBeNil)
+			So(err.Error(), ShouldEqual, "failed as unexpected code from search reindex api: 409")
+			So(apiError.ErrorStatus(err), ShouldEqual, http.StatusConflict)
+
+			Convey("then the expected empty job is returned", func() {
```
cookel2, line 291: This needs to say that the expected task is returned,
```suggestion
			Convey("Then the expected task is returned", func() {
```

sdk/v1/client_test.go
```diff
@@ -208,3 +208,95 @@ func TestClient_PostJob(t *testing.T) {
...
+				So(task, ShouldResemble, models.Task{})
+			})
+
+			Convey("and client.Do should be called once with the expected parameters", func() {
```
cookel2, line 295:
```suggestion
			Convey("And client.Do should be called once with the expected parameters", func() {
```

sdk/v1/client_test.go
```diff
@@ -208,3 +208,95 @@ func TestClient_PostJob(t *testing.T) {
...
+
+			Convey("And client.Do should be called once with the expected parameters", func() {
+				doCalls := httpClient.DoCalls()
+				So(doCalls, ShouldHaveLength, 1)
+				So(doCalls[0].Req.URL.Path, ShouldEqual, pathToCheck)
+			})
+		})
+	})
+	Convey("given a 404 response", t, func() {
+		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusConflict}, nil)
```
cookel2, line 282:
```suggestion
		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusNotFound}, nil)
```

cookel2 (changes requested):
Bhupendra there are a few things to change here if you don't mind. Thanks, Ella

sdk/v1/client.go
bpathak-ons: well spotted....now implemented

cookel2 (approved):
looks good and works for me

sdk/client.go
```diff
@@ -17,6 +17,7 @@ type Client interface {
 	Checker(ctx context.Context, check *health.CheckState) error
 	Health() *healthcheck.Client
 	PostJob(ctx context.Context, headers Headers) (models.Job, error)
+	PostTasksCount(ctx context.Context, headers Headers, jobID string) (models.Task, error)
```
cookel2, line 20: I think it would be better just to call it 'PostTask' otherwise it could get confused with the PUT endpoint that will be used to update the count of tasks. Also, it needs another parameter for the request body e.g. 
```suggestion
	PostTask(ctx context.Context, headers Headers, jobID string, taskToCreate models.TaskToCreate) (models.Task, error)
```

sdk/v1/client.go
```diff
@@ -92,6 +92,31 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (models.
 	return job, nil
 }
 
+// PostTasksCount Updates tasks count for processing
+func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string, payload []byte) (models.Task, error) {
```
cookel2, line 96: I think this should be a TaskToCreate
```suggestion
func (cli *Client) PostTasksCount(ctx context.Context, headers client.Headers, jobID string, taskToCreate models.TaskToCreate) (models.Task, error) {
```

sdk/v1/client.go
```diff
@@ -92,6 +92,31 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (models.
...
+	var task models.Task
+
+	if headers.ServiceAuthToken == "" {
+		headers.ServiceAuthToken = cli.serviceToken
+	}
+
+	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
```
cookel2, line 103: Something like this will be needed here: 

payload, _ := json.Marshal(taskToCreate)

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
 	})
 }
+
+func TestClient_PostTasksCount(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	testJobID := "883c81fd-726d-4ea3-9db8-7e7c781a01cc"
+	pathToCheck := "v1/jobs/883c81fd-726d-4ea3-9db8-7e7c781a01cc/tasks"
+
+	mockTaskToCreate := `{"task_name":"zebedee","number_of_documents": "10"}`
```
cookel2, line 222: This will also need to be a TaskToCreate:
```suggestion
		mockTaskToCreate := models.TaskToCreate{
		TaskName:          "zebedee",
		NumberOfDocuments: 10,
	}
```

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
+			},
+			nil)
+
+		searchReindexClient := newSearchReindexClient(t, httpClient)
+
+		Convey("When search-reindexClient.PostTasksCount is called", func() {
+			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID, testPayload)
+			So(err, ShouldBeNil)
+
+			Convey("Then the expected jobid is returned", func() {
```
cookel2, line 255:
```suggestion
			Convey("Then the expected jobid, task name, and number of documents, are returned", func() {
```

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
+				So(testJobID, ShouldEqual, expectedTask.JobID)
+				So(task.NumberOfDocuments, ShouldEqual, expectedTask.NumberOfDocuments)
```
cookel2, line 257: Please also test the task name is correct:

So(task.TaskName, ShouldEqual, expectedTask.TaskName)

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
+			})
+
+			Convey("And client.Do should be called once with the expected parameters", func() {
+				doCalls := httpClient.DoCalls()
+				So(doCalls, ShouldHaveLength, 1)
+				So(doCalls[0].Req.URL.Path, ShouldEqual, pathToCheck)
```
cookel2, line 263: Please check that the request body is correct:

reqBody, _ := json.Marshal(mockTaskToCreate)
expectedBody := io.NopCloser(bytes.NewReader(reqBody))
So(doCalls[0].Req.Body, ShouldResemble, expectedBody)

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
+			})
+		})
+	})
+	Convey("Given a 500 response", t, func() {
+		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusInternalServerError}, nil)
+		searchReindexClient := newSearchReindexClient(t, httpClient)
+
+		Convey("When search-reindexClient.PostTaskCount is called", func() {
+			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID, testPayload)
```
cookel2, line 272: This will need to be changed then:
```suggestion
			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID, mockTaskToCreate)
```

sdk/v1/client_test.go
```diff
@@ -190,25 +190,120 @@ func TestClient_PostJob(t *testing.T) {
...
+				So(doCalls[0].Req.URL.Path, ShouldEqual, pathToCheck)
+			})
+		})
+	})
+	Convey("Given a 404 response", t, func() {
+		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusNotFound}, nil)
+		searchReindexClient := newSearchReindexClient(t, httpClient)
+
+		Convey("When search-reindexClient.PostTasksCount is called", func() {
+			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID, testPayload)
```
cookel2, line 293:
```suggestion
			task, err := searchReindexClient.PostTasksCount(ctx, client.Headers{}, testJobID, mockTaskToCreate)
```

cookel2 (changes requested):
There are a few more comments from me.

sdk/v1/client.go
bpathak-ons: nope, the client will be invoked with payload as a byte slice. please let me know if need to explain further.

sdk/v1/client_test.go
bpathak-ons: nope. again this is a representation of byte slice.

sdk/v1/client.go
bpathak-ons: nope, see my comments as above

sdk/v1/client_test.go
bpathak-ons: not implemented in line with my other comments as above

sdk/v1/client_test.go
bpathak-ons: not needed, I think, as payload remains the byte slice

PR #71, opened 2022-04-01
Update client instantiation with methods New and NewClientWithHealth

nshumoogum:
### What

See title, tests updated too

### How to review

Check client instantiation makes sense, check for typos

### Who can review

Anyone

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 
+	healthcheck "github.com/ONSdigital/dp-api-clients-go/v2/health"
+	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	dprequest "github.com/ONSdigital/dp-net/v2/request"
 	"github.com/ONSdigital/dp-search-reindex-api/models"
 )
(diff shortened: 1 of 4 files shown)


sdk/client.go
```diff
@@ -5,14 +5,19 @@ import (
...
+	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	dprequest "github.com/ONSdigital/dp-net/v2/request"
 	"github.com/ONSdigital/dp-search-reindex-api/models"
 )
 
 //go:generate moq -out ./mocks/client.go -pkg mocks . Client
 
 type Client interface {
+	Checker(ctx context.Context, check *health.CheckState) error
+	HealthClient() *healthcheck.Client
```
rahulPalliyalil, line 18: (suggestion): I think the interface ```HealthClient``` kind of cause stutter. ```https://go.dev/doc/effective_go```. An effective name would be something like Health(), so the consumers can call ```client.Health()```. rather than ```client.HealthClient```.

sdk/v1/client.go
```diff
@@ -30,28 +29,36 @@ type Client struct {
...
 	}
 }
 
-// NewWithHealthClient creates a new instance of Client,
-// reusing the URL and Clienter from the provided health check client.
-func NewClientWithHealthcheck(serviceToken string, hcCli *healthcheck.Client) *Client {
+// NewWithHealthClient creates a new instance of SearchReindexAPI Client,
+// reusing the URL and Clienter from the provided healthcheck client
+func NewWithHealthClient(serviceToken string, hcCli *healthcheck.Client) *Client {
 	return &Client{
```
rahulPalliyalil, line 44: (minor): I think here you would have to check if the hcCli is nil and if it is nil probably return a nil client and an error ?

rahulPalliyalil (changes requested):

rahulPalliyalil (approved):

PR #72, opened 2022-04-04
Update sdk client for search reindex api

nshumoogum:
### What

Change how we instantiate the client and check for nil client values

### How to review

- Sanity check changes

### Who can review

Anyone

### DIFF
--- a/sdk/v1/client.go
+++ b/sdk/v1/client.go
@@ -43,7 +43,7 @@ func New(searchReindexAPIURL, serviceToken string) *Client {
 // reusing the URL and Clienter from the provided healthcheck client
 func NewWithHealthClient(serviceToken string, hcCli *healthcheck.Client) (*Client, error) {
 	if hcCli == nil {
-		return nil, errors.New("client is nil")
+		return nil, errors.New("health client is nil")
 	}
 	return &Client{
 		apiVersion:   apiVersion,
--- a/sdk/v1/client.go
+++ b/sdk/v1/client.go
@@ -49,13 +49,13 @@ func NewWithHealthClient(serviceToken string, hcCli *healthcheck.Client) *Client
 }
 
 // URL returns the URL used by this client
-func (c *Client) URL() string {
-	return c.hcCli.URL
+func (cli *Client) URL() string {
+	return cli.hcCli.URL
 }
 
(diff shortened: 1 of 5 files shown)


sdk/v1/client.go
```diff
@@ -30,28 +30,39 @@ type Client struct {
...
 }
 
-// NewWithHealthClient creates a new instance of Client,
-// reusing the URL and Clienter from the provided health check client.
-func NewClientWithHealthcheck(serviceToken string, hcCli *healthcheck.Client) *Client {
+// NewWithHealthClient creates a new instance of SearchReindexAPI Client,
+// reusing the URL and Clienter from the provided healthcheck client
+func NewWithHealthClient(serviceToken string, hcCli *healthcheck.Client) (*Client, error) {
+	if hcCli == nil {
+		return nil, errors.New("client is nil")
```
rahulPalliyalil, line 46: (suggestion): would it be better to say health check client is nil ? rather than a generic ```client is nil``` message ?

sdk/v1/client_test.go
```diff
@@ -44,9 +44,13 @@ func newMockHTTPClient(r *http.Response, err error) *dphttp.ClienterMock {
 	}
 }
 
-func newSearchReindexClient(httpClient *dphttp.ClienterMock) *Client {
+func newSearchReindexClient(t *testing.T, httpClient *dphttp.ClienterMock) *Client {
 	healthClient := healthcheck.NewClientWithClienter(serviceName, testHost, httpClient)
```
rahulPalliyalil, line 48: (sugestion): can we run this is parallel ? t.Parallel ?

rahulPalliyalil (approved):

sdk/v1/client_test.go
nshumoogum: No this is not a test, just a function used in the unit tests, the reason for passing *testing.T was to fail the test inside this function instead of handling it in the actual unit tests for every occasion (duplication of code)

sdk/v1/client.go
nshumoogum: Yeah i think your right

rahulPalliyalil (approved):

PR #75, opened 2022-04-05
Extend client to call patch jobs

cookel2:
# What has changed

The Search Reindex Client has been extended to include the following method:

PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) (string, error)

And unit tests have been added to test the following:
- The Search Reindex Client returns 204 and an ETag (the success scenario)
- The Search Reindex Client returns a 400 response (bad patch request)
- The Search Reindex Client returns a 500 response (search reindex error)
- The Search Reindex Client returns a 409 response (conflict with other reindex job)
- The Search Reindex Client returns a 404 response (reindex job not found)

# How to review

The linter can be run using this command: makee lint

# Who can review

Anyone but me

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -45,13 +45,13 @@ func (h *Headers) Add(req *http.Request) {
 	}
 	fmt.Println("the ETag value is: " + h.ETag)
 	fmt.Println("the auth value is: " + h.ServiceAuthToken)
+	fmt.Println("the IfMatch value is: " + h.IfMatch)
 	if h.ETag != "" {
 		dpclients.SetETag(req, h.ETag)
 	}
 
 	if h.IfMatch != "" {
-		// TODO Set IfMatch header
-		fmt.Println("currently not handling IfMatch header")
+		dpclients.SetIfMatch(req, h.IfMatch)
 	}
 
 	if h.ServiceAuthToken != "" {
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -17,7 +17,7 @@ type Client interface {
 	Checker(ctx context.Context, check *health.CheckState) error
 	Health() *healthcheck.Client
 	PostJob(ctx context.Context, headers Headers) (models.Job, error)
-	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) error
+	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) (string, error)
 	URL() string
 }
 
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -13,7 +13,7 @@ import (
 
 type Client interface {
 	PostJob(ctx context.Context, headers Headers) (models.Job, error)
-	PatchJob(ctx context.Context, headers Headers, jobID string, body PatchOpsList) error
+	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) error
 }
 
 type Headers struct {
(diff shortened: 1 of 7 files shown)


sdk/client.go
```diff
@@ -33,19 +34,29 @@ type Options struct {
...
+type PatchOperation struct {
+	Op    string
+	Path  string
+	Value interface{}
+}
+
+func (h *Headers) Add(req *http.Request) error {
 	if h == nil {
-		return
+		return nil
```
bpathak-ons, line 45: worth adding log info for missing header

sdk/client.go
```diff
@@ -33,19 +34,29 @@ type Options struct {
...
 	}
 
 	if h.ETag != "" {
-		// TODO Set ETag header
-		fmt.Println("currently not handling ETag header")
+		err := dpclients.SetETag(req, h.ETag)
+		if err != nil {
+			return err
```
bpathak-ons, line 51: log error with err details pls

sdk/client.go
```diff
@@ -33,19 +34,29 @@ type Options struct {
...
+		}
 	}
 
 	if h.IfMatch != "" {
-		// TODO Set IfMatch header
-		fmt.Println("currently not handling IfMatch header")
+		err := dpclients.SetIfMatch(req, h.IfMatch)
+		if err != nil {
+			return err
```
bpathak-ons, line 58: same log err details pls

sdk/client_test.go
```diff
@@ -8,15 +8,19 @@ import (
...
+	eTagHeader    = "ETag"
+	ifMatchHeader = "If-Match"
 )
 
 func TestHeaders_Add(t *testing.T) {
 	t.Parallel()
 
 	Convey("Given the sdk Headers struct contains a value for ETag", t, func() {
-		req := &http.Request{}
+		req := &http.Request{
```
bpathak-ons, line 21: can this be moved at the top of the test to avoid repetition of code?

sdk/v1/client.go
```diff
@@ -114,17 +134,23 @@ func (cli *Client) callReindexAPI(ctx context.Context, path, method string, head
...
 			Err:  fmt.Errorf("failed to create request for call to search reindex api, error is: %v", err),
 			Code: http.StatusInternalServerError,
 		}
 	}
 
-	headers.Add(req)
+	err = headers.Add(req)
+	if err != nil {
+		return nil, nil, apiError.StatusError{
+			Err:  fmt.Errorf("failed to add headers to request, error is: %v", err),
```
bpathak-ons, line 146: log rather than fmt, I think

sdk/client.go
cookel2: ok, done

sdk/client.go
cookel2: ok, done

sdk/client.go
cookel2: ok, done

sdk/client_test.go
cookel2: good idea, done

sdk/v1/client.go
cookel2: I don't think I can use log here as it would return a string but it needs to be a value for the Err in the apiError.StatusError struct. I would prefer to leave it as it is. This code is only changed because I added a nil to the values being returned in line 145.

bpathak-ons (approved):
looks ok

PR #80, opened 2022-04-14
Task model updates for client with pass by reference

bpathak-ons:
### What

* Task model changes using pass by reference  instead of pass by value

### How to review

* code sense check
* CI pass

### Who can review

* Anyone

### DIFF
--- a/sdk/v1/client_test.go
+++ b/sdk/v1/client_test.go
@@ -269,9 +269,9 @@ func TestClient_PostTasksCount(t *testing.T) {
 				So(task.NumberOfDocuments, ShouldEqual, expectedTask.NumberOfDocuments)
 			})
 
-			Convey("And an ETag is returned", func() {
+			Convey("And an responseheader is returned", func() {
 				So(respHeaders, ShouldNotBeNil)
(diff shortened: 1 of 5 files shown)


sdk/v1/client_test.go
```diff
@@ -316,14 +321,19 @@ func TestClient_PostTasksCount(t *testing.T) {
 		searchReindexClient := newSearchReindexClient(t, httpClient)
 
 		Convey("When search-reindexClient.PostTasksCount is called", func() {
-			respETag, task, err := searchReindexClient.PostTasksCount(ctx, headers, testJobID, testPayload)
+			respHeaders, task, err := searchReindexClient.PostTasksCount(ctx, headers, testJobID, testPayload)
```
cookel2, line 324: This is not really in this PR but please would you mind changing line 319 to be 404 instead of 409:

			Convey("Given a 404 response", t, func() {

cookel2 (commented):
Please see comment, thanks.

cookel2 (approved):
I tested it locally and it looks good.

PR #89, opened 2022-05-03
Fix path and refactor post tasks count

cookel2:
# What has changed

The PostTasksCount method was giving the following error when used:

unsupported protocol scheme \“\”

These changes fix the problem, which was with the path, and also refactor the PostTasksCount method and rename it PostTask.

# How to review

Check that the code looks correct and that the following tests pass:

make test

go test -component

make lint

# Who can review

Anyone but me

### DIFF
--- a/sdk/v1/client.go
+++ b/sdk/v1/client.go
@@ -90,13 +90,15 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (*models
 	return &job, nil
 }
 
-// PostTasksCount Updates tasks count for processing
-func (cli *Client) PostTasksCount(ctx context.Context, reqheaders client.Headers, jobID string, payload []byte) (*client.RespHeaders, *models.Task, error) {
+// PostTask creates or updates a task, for the job with id = jobID, containing the number of documents to be processed
+func (cli *Client) PostTask(ctx context.Context, reqheaders client.Headers, jobID string, taskToCreate models.TaskToCreate) (*client.RespHeaders, *models.Task, error) {
 	if reqheaders.ServiceAuthToken == "" {
 		reqheaders.ServiceAuthToken = cli.serviceToken
 	}
 
-	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
+// 	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
+	path := cli.hcCli.URL + "/" + cli.apiVersion + jobsEndpoint + "/" + jobID + "/tasks"
+	payload, _ := json.Marshal(taskToCreate)
 
 	respHeader, b, err := cli.callReindexAPI(ctx, path, http.MethodPost, reqheaders, payload)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


sdk/v1/client.go
```diff
@@ -90,13 +90,14 @@ func (cli *Client) PostJob(ctx context.Context, headers client.Headers) (*models
...
-func (cli *Client) PostTasksCount(ctx context.Context, reqheaders client.Headers, jobID string, payload []byte) (*client.RespHeaders, *models.Task, error) {
+// PostTask creates or updates a task, for the job with id = jobID, containing the number of documents to be processed
+func (cli *Client) PostTask(ctx context.Context, reqheaders client.Headers, jobID string, taskToCreate models.TaskToCreate) (*client.RespHeaders, *models.Task, error) {
 	if reqheaders.ServiceAuthToken == "" {
 		reqheaders.ServiceAuthToken = cli.serviceToken
 	}
 
-	path := fmt.Sprintf("%s/jobs/%s/tasks", cli.apiVersion, jobID)
+	path := cli.hcCli.URL + "/" + cli.apiVersion + jobsEndpoint + "/" + jobID + "/tasks"
+	payload, _ := json.Marshal(taskToCreate)
```
nshumoogum, line 100: We should handle the error here

sdk/v1/client_test.go
```diff
@@ -229,13 +229,20 @@ func TestClient_PostJob(t *testing.T) {
...
-	mockTaskToCreate := `{"task_name":"zebedee","number_of_documents": "10"}`
-	testPayload := []byte(mockTaskToCreate)
+	taskToCreate := models.TaskToCreate{
+		TaskName:          "zebedee",
+		NumberOfDocuments: 10,
+	}
+
+	testPayload, _ := json.Marshal(taskToCreate)
+	reqBody := bytes.NewReader(testPayload)
+	req, _ := http.NewRequest("POST", pathToCheck, reqBody)
```
nshumoogum, line 244: We should check the error here

sdk/v1/client_test.go
```diff
@@ -229,13 +229,20 @@ func TestClient_PostJob(t *testing.T) {
...
+	pathToCheck := "/v1/jobs/883c81fd-726d-4ea3-9db8-7e7c781a01cc/tasks"
 
-	mockTaskToCreate := `{"task_name":"zebedee","number_of_documents": "10"}`
-	testPayload := []byte(mockTaskToCreate)
+	taskToCreate := models.TaskToCreate{
+		TaskName:          "zebedee",
+		NumberOfDocuments: 10,
+	}
+
+	testPayload, _ := json.Marshal(taskToCreate)
```
nshumoogum, line 242: Check for error

nshumoogum (changes requested):

sdk/v1/client.go
cookel2: done

sdk/v1/client_test.go
cookel2: done

sdk/v1/client_test.go
cookel2: done

nshumoogum (approved):

PR #91, opened 2022-05-05
Extend client to get tasks

bpathak-ons:
### What

* Client for getTasks to get number to tasks assigned to a job

### How to review

* code sense check
* CI pass

### Who can review

* Anyone

### DIFF
--- a/sdk/v1/client_test.go
+++ b/sdk/v1/client_test.go
@@ -58,7 +58,7 @@ var (
 		ETag:              `"76b6890f1321590998d5fd8d293b620581ff3541"`,
 	}
 
-	exectedTasks = models.Tasks{
+	expectedTasks = models.Tasks{
 		Count:      2,
 		TaskList:   []models.Task{expectedTask, expectedTask2},
 		Limit:      5,
@@ -711,7 +711,7 @@ func TestClient_GetTasks(t *testing.T) {
 	}
 
 	Convey("Given clienter.Do doesn't return an error", t, func() {
-		body, err := json.Marshal(exectedTasks)
+		body, err := json.Marshal(expectedTasks)
 		if err != nil {
 			t.Errorf("failed to setup test data, error: %v", err)
 		}
--- a/sdk/v1/client_test.go
+++ b/sdk/v1/client_test.go
@@ -43,7 +43,6 @@ var (
 		},
 		NumberOfDocuments: 10,
 		TaskName:          "zebedee",
-		ETag:              `"56b6890f1321590998d5fd8d293b620581ff3531"`,
 	}
 
 	expectedTask2 = models.Task{
@@ -55,7 +54,6 @@ var (
 		},
 		NumberOfDocuments: 20,
 		TaskName:          "dataset-api",
-		ETag:              `"76b6890f1321590998d5fd8d293b620581ff3541"`,
 	}
 
 	expectedTasks = models.Tasks{
@@ -607,7 +605,7 @@ func TestClient_GetTask(t *testing.T) {
 				StatusCode: http.StatusCreated,
 				Body:       io.NopCloser(bytes.NewReader(body)),
 				Header: http.Header{
-					"Etag": []string{expectedTask.ETag},
+					"Etag": []string{testETag},
 				},
 			},
 			nil)
@@ -626,7 +624,7 @@ func TestClient_GetTask(t *testing.T) {
 
 			Convey("And an ETag is returned", func() {
 				So(respHeaders, ShouldNotBeNil)
-				So(respHeaders, ShouldResemble, &client.RespHeaders{ETag: expectedTask.ETag})
+				So(respHeaders, ShouldResemble, &client.RespHeaders{ETag: testETag})
 			})
 
 			Convey("And client.Do should be called once with the expected parameters", func() {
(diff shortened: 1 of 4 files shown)


sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+				So(tasks.TotalCount, ShouldEqual, 20)
+				So(tasks.TaskList[0].JobID, ShouldEqual, "883c81fd-726d-4ea3-9db8-7e7c781a01cc")
+				So(tasks.TaskList[0].TaskName, ShouldEqual, "zebedee")
+				So(tasks.TaskList[0].NumberOfDocuments, ShouldEqual, 10)
+				So(tasks.TaskList[1].JobID, ShouldEqual, "883c81fd-726d-4ea3-9db8-7e7c781a01cc")
+				So(tasks.TaskList[1].TaskName, ShouldEqual, "dataset-api")
+				So(tasks.TaskList[1].NumberOfDocuments, ShouldEqual, 20)
+			})
+
+			Convey("And an ETag is returned", func() {
```
cookel2, line 749: Although an ETag is being returned by the test there is no ETag currently being returned, in the response header, if the GET tasks endpoint is actually called. So I think this part of the test should be removed as it is not currently doing this.

sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+			So(task, ShouldBeNil)
+
+			So(err.Error(), ShouldEqual, "failed as unexpected code from search reindex api: 500")
+			So(apiError.ErrorStatus(err), ShouldEqual, http.StatusInternalServerError)
+
+			Convey("Then the expected empty task is returned", func() {
+				So(task, ShouldBeNil)
+			})
+
+			Convey("And an empty ETag is returned", func() {
```
cookel2, line 782: Similarly I think this part should be removed as not even an empty ETag is currently being returned.

sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+			So(err, ShouldNotBeNil)
+			So(task, ShouldBeNil)
+			So(err.Error(), ShouldEqual, "failed as unexpected code from search reindex api: 404")
+			So(apiError.ErrorStatus(err), ShouldEqual, http.StatusNotFound)
+
+			Convey("Then the expected empty task is returned", func() {
+				So(task, ShouldBeNil)
+			})
+
+			Convey("Then an empty ETag is returned", func() {
```
cookel2, line 811: Same here.

cookel2 (changes requested):
Just a few changes requested please.

sdk/v1/client_test.go
bpathak-ons: correct, trello task for generating ETag in response header is - https://trello.com/c/flHAokOt/1152-search-reindex-api-ifmatch-and-etag-headers-for-get-jobs-jobid-tasks-endpoint

nshumoogum: I would right the client and unit test based on the swagger spec (assuming this is up to date)

In this case as Bhupendra says we will want to check the ETag is in the response header and not returned in the body. the data is mocked and so we can remove the ETag entry in the expectedTask variable when all the client methods have been updated.

See my next comment on removing duplicate assertions based on the above

sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+				So(tasks.TaskList[0].JobID, ShouldEqual, "883c81fd-726d-4ea3-9db8-7e7c781a01cc")
+				So(tasks.TaskList[0].TaskName, ShouldEqual, "zebedee")
+				So(tasks.TaskList[0].NumberOfDocuments, ShouldEqual, 10)
+				So(tasks.TaskList[1].JobID, ShouldEqual, "883c81fd-726d-4ea3-9db8-7e7c781a01cc")
+				So(tasks.TaskList[1].TaskName, ShouldEqual, "dataset-api")
+				So(tasks.TaskList[1].NumberOfDocuments, ShouldEqual, 20)
+			})
+
+			Convey("And an ETag is returned", func() {
+				So(respHeaders, ShouldNotBeNil)
```
nshumoogum, line 750: These two assertions (line 750 and 751) can be removed as they are duplicating the same checks on ETag entry

```suggestion
```

sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+			So(task, ShouldBeNil)
+
+			So(err.Error(), ShouldEqual, "failed as unexpected code from search reindex api: 500")
+			So(apiError.ErrorStatus(err), ShouldEqual, http.StatusInternalServerError)
+
+			Convey("Then the expected empty task is returned", func() {
+				So(task, ShouldBeNil)
+			})
+
+			Convey("And an empty ETag is returned", func() {
```
nshumoogum, line 782: Probably want to update the message in Convey:
```suggestion
			Convey("And no headers are returned", func() {
```

sdk/v1/client_test.go
```diff
@@ -679,3 +699,127 @@ func TestClient_GetTask(t *testing.T) {
...
+			So(err, ShouldNotBeNil)
+			So(task, ShouldBeNil)
+			So(err.Error(), ShouldEqual, "failed as unexpected code from search reindex api: 404")
+			So(apiError.ErrorStatus(err), ShouldEqual, http.StatusNotFound)
+
+			Convey("Then the expected empty task is returned", func() {
+				So(task, ShouldBeNil)
+			})
+
+			Convey("Then an empty ETag is returned", func() {
```
nshumoogum, line 811: Again probably want to update message in Convey func:

```suggestion
			Convey("Then no headers are returned", func() {
```

sdk/v1/client_test.go
bpathak-ons: ETag entry removed with duplicate assertion.

nshumoogum (approved):

cookel2 (approved):
looks good to me

PR #94, opened 2022-05-09
Extend client to get jobs endpoint

bpathak-ons:
### What

* Client for getJobs to get list of jobs
* Include query parameters, offset, limit and others by using the Options struct in method signature
* unit test

### How to review

* code sense check
* CI pass

### Who can review

* Anyone

### DIFF
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -21,6 +21,7 @@ type Client interface {
 	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) (*RespHeaders, error)
 	PostTask(ctx context.Context, headers Headers, jobID string, taskToCreate models.TaskToCreate) (*RespHeaders, *models.Task, error)
 	GetTask(ctx context.Context, headers Headers, jobID, taskName string) (*RespHeaders, *models.Task, error)
+	GetJobs(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
 	URL() string
 }
 
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -20,9 +20,9 @@ type Client interface {
 	PostJob(ctx context.Context, headers Headers) (*models.Job, error)
 	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) (*RespHeaders, error)
 	PostTask(ctx context.Context, headers Headers, jobID string, taskToCreate models.TaskToCreate) (*RespHeaders, *models.Task, error)
-	GetTask(ctx context.Context, headers Headers, jobID, taskName string) (*RespHeaders, *models.Task, error)
-	GetJobs(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
 	GetJob(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
+	GetJobs(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
+	GetTask(ctx context.Context, headers Headers, jobID, taskName string) (*RespHeaders, *models.Task, error)
 	GetTasks(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Tasks, error)
 	URL() string
 }
--- a/sdk/client.go
+++ b/sdk/client.go
@@ -9,7 +9,6 @@ import (
 	health "github.com/ONSdigital/dp-healthcheck/healthcheck"
 	dprequest "github.com/ONSdigital/dp-net/v2/request"
 	"github.com/ONSdigital/dp-search-reindex-api/models"
-	"github.com/ONSdigital/dp-search-reindex-api/mongo"
 	"github.com/ONSdigital/log.go/v2/log"
 )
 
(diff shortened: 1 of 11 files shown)


sdk/client.go
```diff
@@ -20,8 +20,9 @@ type Client interface {
 	PostJob(ctx context.Context, headers Headers) (*models.Job, error)
 	PatchJob(ctx context.Context, headers Headers, jobID string, body []PatchOperation) (*RespHeaders, error)
 	PostTask(ctx context.Context, headers Headers, jobID string, taskToCreate models.TaskToCreate) (*RespHeaders, *models.Task, error)
-	GetTask(ctx context.Context, headers Headers, jobID, taskName string) (*RespHeaders, *models.Task, error)
 	GetJob(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
+	GetJobs(ctx context.Context, reqheader Headers, jobID string) (*RespHeaders, *models.Job, error)
```
cookel2, line 24: This needs to pass in the Options:
```suggestion
	GetJobs(ctx context.Context, reqheader Headers, jobID string, options Options) (*RespHeaders, *models.Job, error)
```

sdk/v1/client.go
```diff
@@ -229,6 +229,34 @@ func (cli *Client) GetJob(ctx context.Context, reqheader client.Headers, jobID s
 	return &respHeaders, &job, nil
 }
 
+func (cli *Client) GetJobs(ctx context.Context, reqheader client.Headers) (*client.RespHeaders, *models.Jobs, error) {
```
cookel2, line 232: Here too:
```suggestion
func (cli *Client) GetJobs(ctx context.Context, reqheader client.Headers, options client.Options) (*client.RespHeaders, *models.Jobs, error) {
```

sdk/v1/client.go
```diff
@@ -229,6 +229,34 @@ func (cli *Client) GetJob(ctx context.Context, reqheader client.Headers, jobID s
...
+	if reqheader.ServiceAuthToken == "" {
+		reqheader.ServiceAuthToken = cli.serviceToken
+	}
+
+	path := fmt.Sprintf("%s/%s/jobs", cli.hcCli.URL, cli.apiVersion)
```
cookel2, line 237: The options will each need to be added to the path if they are present. It needs to end up looking something like this: http://localhost:25700/jobs?offset=1&limit=7&sort=last_updated

cookel2 (changes requested):
The options need to be added to the GetJobs function.

sdk/v1/client_test.go
```diff
@@ -810,7 +832,135 @@ func TestClient_GetTasks(t *testing.T) {
...
+				expectedIfMatchHeader[0] = "*"
+				So(doCalls[0].Req.Header[ifMatchHeader], ShouldResemble, expectedIfMatchHeader)
+			})
+		})
+	})
+	Convey("Given a 500 response", t, func() {
+		httpClient := newMockHTTPClient(&http.Response{StatusCode: http.StatusInternalServerError}, nil)
+		searchReindexClient := newSearchReindexClient(t, httpClient)
+
+		Convey("When search-reindexClient.GetTask is called", func() {
```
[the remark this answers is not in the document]
cookel2: I still see GetTask on this line.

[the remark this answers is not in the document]
cookel2: looks good

sdk/v1/client.go
bpathak-ons: handled options for the endpoint now

sdk/v1/client.go
```diff
@@ -233,6 +235,45 @@ func (cli *Client) GetJob(ctx context.Context, reqheader client.Headers, jobID s
...
+
+	configValues, _ := config.Get()
+	validOffset, err := cli.ValidateOptions(options.Offset)
+	if err != nil || validOffset == "0" {
+		validOffset = strconv.Itoa(configValues.DefaultOffset)
+	}
+
+	validLimit, err := cli.ValidateOptions(options.Limit)
+	if err != nil || validLimit == "0" {
+		validLimit = strconv.Itoa(configValues.DefaultLimit)
```
cookel2, line 251: If someone has specified a limit of 0 then this would cause the endpoint to have no limit and therefore would list all the jobs even if there were more than the maximum that should be returned. So, to stop this from happening we could use the default limit as you have done. However, I think it might be better to use the default maximum limit.
```suggestion
		validLimit = strconv.Itoa(configValues.DefaultMaxLimit)
```

cookel2 (changes requested):
just a couple of things please

sdk/v1/client.go
```diff
@@ -233,6 +235,45 @@ func (cli *Client) GetJob(ctx context.Context, reqheader client.Headers, jobID s
...
+
+	configValues, _ := config.Get()
+	validOffset, err := cli.ValidateOptions(options.Offset)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	validLimit, err := cli.ValidateOptions(options.Limit)
+	if err != nil {
+		validLimit = strconv.Itoa(configValues.DefaultMaxLimit)
```
cookel2, line 251: Sorry Bhupendra but on second thoughts I think we need to return the error here too:
```suggestion
		return nil, nil, err
```

cookel2 (approved):
looks good

PR #96, opened 2022-05-17
Refactor GetJob and GetJobs handlers

cookel2:
# What has changed

GetJobHandler
- Pass existing context into line 144 (reuse variable ctx created when handler is first called)
- Do error handling before setting the content-type header and status header (otherwise if this fails then we will return an error message instead of json in the response body which will be of content-type: plain/text)

GetsJobsHandler
- A long log.Info line on line 179 in api/jobs.go is now reduced to be more concise: "GetJobsHandler: returns a list of Job resources"

# How to review

Check that the code looks correct. Also call the GetJob and GetJobs endpoints:
Need to run:
- docker-compose up -d (for Mongo, ES, and kafka)
- vault server -dev
- zebedee (./run.sh)
- search api (with ES port 11200)
- search reindex api

Then run the following curl commands:
To create a new job
curl -X POST http://localhost:25700/jobs | jq

To get a particular job:
curl -s http://localhost:25700/jobs/<job id from created job> | jq

To get a list of jobs:
curl -s http://localhost:25700/jobs | jq
curl -s 'http://localhost:25700/jobs?offset=2&limit=9' | jq

The linter can be run using this command: make lint

# Who can review

Anyone but me

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -175,7 +175,7 @@ func (api *API) GetJobHandler(w http.ResponseWriter, req *http.Request) {
 // last_updated time (ascending).
 func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	ctx := req.Context()
-	log.Info(ctx, "Entering handler function, which calls GetJobs and returns a list of existing Job resources held in the JobStore.")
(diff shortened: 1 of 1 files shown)


Marianasilvago (approved):
LGTM

PR #99, opened 2022-05-18
simplify component tests

justinpjose:
### What
- Remove the step `the search api is working correctly` in all the other features except for the `posting_a_job.feature`
- This is because the Search API is only used in the `CreateJob` handler and not in any of the others
- This was achieved by using the mongo client within the component test rather than doing a `POST /jobs` request within the component test to generate jobs in the datastore (`mongo`). 
- This makes it more performant as it is now only doing the create job in `mongo` when asked whereas the component test would have previously done the `POST /jobs` which would do additional processes such as creating an index and getting the index name which was not needed
- This also simplifies and removes code that reads the HTTP response body from those calls which is no longer required 

### How to review
- Check if the tests pass

### Who can review
Anyone

### DIFF
--- a/features/latest_version_features/getting_a_job.feature
+++ b/features/latest_version_features/getting_a_job.feature
@@ -2,7 +2,7 @@ Feature: Getting a job
 
   Scenario: Job exists in the Job Store and a get request returns it successfully
 
-    Given set the api version to undefined for incoming requests
+    Given the api version is undefined for incoming requests
     And I have generated 1 jobs in the Job Store
     When I call GET /jobs/{id} using the generated id
     Then the response should contain values that have these structures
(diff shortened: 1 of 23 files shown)


cookel2 (commented):
Just one suggestion but see what you think.

features/latest_version_features/getting_a_job.feature
```diff
@@ -2,8 +2,7 @@ Feature: Getting a job
 
   Scenario: Job exists in the Job Store and a get request returns it successfully
 
-    Given the search api is working correctly
-    And set the api version to undefined for incoming requests
+    Given set the api version to undefined for incoming requests
```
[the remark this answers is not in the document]
justinpjose: That's a good suggestion. I thought of doing this in multiple places (not only for this step but for others as well). I will do this in this PR then

features/latest_version_features/posting_a_job.feature
```diff
@@ -3,11 +3,12 @@ Feature: Posting a job
   Scenario: Job is posted successfully
 
     Given the search api is working correctly
-    And set the api version to undefined for incoming requests
+    Given the api version is undefined for incoming requests
```
gilesburdett, line 6: How does the `Given` on line 6 interact with the `Given` on line 5?

features/v1_features/latest_version_features/posting_a_job.feature
```diff
@@ -3,32 +3,32 @@ Feature: Posting a job
   Scenario: Job is posted successfully
 
     Given the search api is working correctly
-    And set the api version to v1 for incoming requests
+    Given the api version is v1 for incoming requests
```
gilesburdett, line 6: A second instance of the double `Given`

features/latest_version_features/posting_a_job.feature
justinpjose: Good spot. Addressed

features/v1_features/latest_version_features/posting_a_job.feature
justinpjose: Good spot. Addressed

features/latest_version_features/getting_a_job.feature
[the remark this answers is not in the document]
justinpjose: Done

cookel2 (approved):
looks good to me

PR #100, opened 2022-05-20
Add errors to apierrors package

cookel2:
# What has changed

All API errors are now created in the to apierrors package:
Any API errors (i.e. any errors that contain the plain text message returned to the API user) that were not in the apierrors package have now been moved to there.

# How to review

Check that the code looks correct; check that the error messages in the apierrors package are user friendly and do not contain any unnecessary information about how the code is written.

The linter can be run using this command: make lint

# Who can review

Anyone but me


features/v1_features/latest_version_features/patch_job_state_fail.feature
```diff
@@ -62,7 +62,7 @@ Feature: Patch job state - Failure
     Then the HTTP status code should be "304"
     And I should receive the following response:
     """
-      new eTag is same as existing eTag
+      new etag is same as existing etag
```
justinpjose, line 65: When I think about this, it makes more sense to say `no modification made on resource`

This makes more sense as the user knows that no changes were made whereas `new eTag is same as existing eTag` doesn't inform that. Also, the error is generic in which this error can be used for jobs and tasks

cookel2: Yes, that makes sense, done.

justinpjose (approved):

PR #104, opened 2022-05-23
refactor `CreateJob` data layer

justinpjose:
### What
- Refactor `CreateJob` handler data layer so that the validation of data exists in the handler layer
- Simplify `CheckIfProgressJob` by only returning one job which is in state `in-progress` and with the most recent `reindex_started` time rather than returning all the jobs and then iterating through it - this is more performant and does the same job

### How to review
- Check if the changes make sense
- Check if all the tests pass

### Who can review
- Anyone

### DIFF
--- a/mongo/jobs.go
+++ b/mongo/jobs.go
@@ -25,104 +25,81 @@ func (m *JobStore) UnlockJob(ctx context.Context, lockID string) {
 	log.Info(ctx, "job lockID has unlocked successfully")
 }
 
-// CheckNewReindexCanBeCreated checks if a new reindex job can be created depending on if a reindex job is currently in progress between cfg.MaxReindexJobRuntime before now and now
+// CheckInProgressJob checks if a new reindex job can be created depending on if a reindex job is currently in progress between cfg.MaxReindexJobRuntime before now and now
 // It returns an error if a reindex job already exists which is in_progress state
-func (m *JobStore) CheckNewReindexCanBeCreated(ctx context.Context) error {
-	log.Info(ctx, "checking if a new reindex job can be created")
+func (m *JobStore) CheckInProgressJob(ctx context.Context) error {
+	log.Info(ctx, "checking if an existing reindex job is in progress")
 
 	s := m.Session.Copy()
 	defer s.Close()
 
-	// get all the jobs with state "in-progress" and order them by "-reindex_started" starting with most recent
-	iter := s.DB(m.Database).C(m.JobsCollection).Find(bson.M{"state": "in-progress"}).Sort("-reindex_started").Iter()
+	var job models.Job
+
+	// get job with state "in-progress" and the most recent "reindex_started" time
+	err := s.DB(m.Database).C(m.JobsCollection).Find(bson.M{"state": "in-progress"}).Sort("-reindex_started").One(&job)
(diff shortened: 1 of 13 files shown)


mongo/jobs.go
```diff
@@ -25,104 +25,81 @@ func (m *JobStore) UnlockJob(ctx context.Context, lockID string) {
...
 	}
 
-	defer func() {
-		if err := iter.Close(); err != nil {
-			log.Error(ctx, "error closing iterator", err)
+	// checkFromTime is the time of configured variable "MaxReindexJobRuntime" from now
+	checkFromTime := time.Now().Add(-1 * m.cfg.MaxReindexJobRuntime)
+
+	// check if start time of the job in progress is later than the checkFromTime but earlier than now
+	if job.ReindexStarted.After(checkFromTime) && job.ReindexStarted.Before(time.Now()) {
```
nshumoogum, line 53: I wonder if this could be added to the query to mongo as a filter option?

mongo/jobs.go
```diff
@@ -25,104 +25,81 @@ func (m *JobStore) UnlockJob(ctx context.Context, lockID string) {
...
+		return err
 	}
 
-	return newJob, err
+	return nil
 }
 
-// validateJobIDUnique checks if there exists a reindex job in mongo with the given id
-func (m *JobStore) validateJobIDUnique(ctx context.Context, id string) error {
+func (m *JobStore) findJob(ctx context.Context, jobID string) (models.Job, error) {
```
nshumoogum, line 84: Use pointer for `models.Job`

nshumoogum: And update job variable below (line 88) and if there is an error then we can return nil instead of creating an object that is empty an unused by the handler

mongo/jobs.go
```diff
@@ -241,3 +197,30 @@ func (m *JobStore) UpdateJob(ctx context.Context, id string, updates bson.M) err
 
 	return nil
 }
+
+// ValidateJobIDUnique checks if there exists a reindex job in mongo with the given id
+func (m *JobStore) ValidateJobIDUnique(ctx context.Context, id string) error {
```
nshumoogum, line 202: Why was this created? The generation of ids for documents will be unique we chose the function so it's on us; by adding this in we are adding latency to the system for no added value

nshumoogum: Also we are not locking the collection down, so who is to say that once this returns a request to the database does not come in to create a document with the same unique id this validation process is using?

I guess the lock will actually exist in the calling method but still i dont see it.

Anyway to make life simpler i would say this method is not needed, mongo will complain when creating a resource if the id is not unique

nshumoogum (changes requested):

mongo/jobs.go
justinpjose: I have addressed this in https://github.com/ONSdigital/dp-search-reindex-api/pull/105 already

mongo/jobs.go
justinpjose: Done

mongo/jobs.go
justinpjose: Done

nshumoogum (approved):

PR #106, opened 2022-05-24
refactor `GetJobsHandler` data layer

justinpjose:
### What
- Separate getting jobs count functionality in separate function `getJobsCount`
- Return pointer to `models.Jobs`

### How to review
- Check if all the test passes

### Who can review
- Anyone

### DIFF
--- a/api/jobs.go
+++ b/api/jobs.go
@@ -185,6 +185,7 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 	limitParam := req.URL.Query().Get("limit")
 	logData := log.Data{}
 
+	// initialise pagination
 	offset, limit, err := pagination.InitialisePagination(api.cfg, offsetParam, limitParam)
 	if err != nil {
 		logData["offset_parameter"] = offsetParam
@@ -199,6 +200,8 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 		Offset: offset,
 		Limit:  limit,
 	}
+
+	// get jobs from mongo
 	jobs, err := api.dataStore.GetJobs(ctx, options)
 	if err != nil {
 		logData["options"] = options
@@ -207,11 +210,13 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// update links for json response
 	for i := range jobs.JobList {
 		jobs.JobList[i].Links.Self = fmt.Sprintf("%s/%s%s", host, v1, jobs.JobList[i].Links.Self)
 		jobs.JobList[i].Links.Tasks = fmt.Sprintf("%s/%s%s", host, v1, jobs.JobList[i].Links.Tasks)
 	}
 
+	// write response
 	err = dpresponse.WriteJSON(w, jobs, http.StatusOK)
 	if err != nil {
 		logData["jobs"] = jobs
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -518,7 +518,7 @@ func TestGetJobsHandler(t *testing.T) {
 
 	Convey("Given a Search Reindex Job API that returns a list of jobs", t, func() {
 		dataStorerMock := &apiMock.DataStorerMock{
-			GetJobsFunc: func(ctx context.Context, options mongo.Options) (models.Jobs, error) {
+			GetJobsFunc: func(ctx context.Context, options mongo.Options) (*models.Jobs, error) {
 				jobs := models.Jobs{}
 				jobsList := make([]models.Job, 2)
 				offsetJobsList := make([]models.Job, 1)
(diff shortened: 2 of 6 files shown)


api/jobs.go
```diff
@@ -199,6 +200,8 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 		Offset: offset,
 		Limit:  limit,
 	}
+
```
rahulPalliyalil, line 203: (suggestion): extra space

api/jobs.go
```diff
@@ -207,11 +210,13 @@ func (api *API) GetJobsHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	// update links for json response
 	for i := range jobs.JobList {
 		jobs.JobList[i].Links.Self = fmt.Sprintf("%s/%s%s", host, v1, jobs.JobList[i].Links.Self)
 		jobs.JobList[i].Links.Tasks = fmt.Sprintf("%s/%s%s", host, v1, jobs.JobList[i].Links.Tasks)
 	}
 
+	// write response
```
rahulPalliyalil, line 219: (suggestion): not sure if you need this comment, as the api is clear on its purpose.

api/jobs_test.go
```diff
@@ -518,7 +518,7 @@ func TestGetJobsHandler(t *testing.T) {
 
 	Convey("Given a Search Reindex Job API that returns a list of jobs", t, func() {
 		dataStorerMock := &apiMock.DataStorerMock{
-			GetJobsFunc: func(ctx context.Context, options mongo.Options) (models.Jobs, error) {
+			GetJobsFunc: func(ctx context.Context, options mongo.Options) (*models.Jobs, error) {
```
rahulPalliyalil, line 521: (suggestion): can we run these in parallel ? ```t.Parallel()```

mongo/jobs.go
```diff
@@ -122,52 +122,67 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (*models.Job, error) {
...
-		results.JobList = make([]models.Job, 0)
-		return results, nil
-	}
+	s := m.Session.Copy()
+	defer s.Close()
 
+	// get all jobs from mongo
 	jobsQuery := s.DB(m.Database).C(m.JobsCollection).Find(bson.M{}).Skip(option.Offset).Limit(option.Limit).Sort("last_updated")
 
+	// populate jobsList from jobsQuery
```
rahulPalliyalil, line 142: (suggestion): again not sure if you need comments here.

mongo/jobs.go
```diff
@@ -122,52 +122,67 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (*models.Job, error) {
...
 	return jobs, nil
 }
 
+// getJobsCount returns the total number of jobs stored in the jobs collection in mongo
+func (m *JobStore) getJobsCount(ctx context.Context) (int, error) {
+	s := m.Session.Copy()
+	defer s.Close()
+
+	logData := log.Data{}
+
```
rahulPalliyalil, line 166: (suggestion): extra space ?

mongo/jobs.go
```diff
@@ -122,52 +122,67 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (*models.Job, error) {
...
+		log.Error(ctx, "error counting jobs", err, logData)
+		return 0, err
+	}
+
+	logData["no_of_jobs"] = numJobs
+	log.Info(ctx, "number of jobs found in jobs collection", logData)
+
+	if numJobs == 0 {
+		log.Info(ctx, "there are no jobs in the data store", logData)
```
rahulPalliyalil, line 180: (suggestion): may be a warning here is better than an info ? ```log.Warn```

mongo/jobs.go
```diff
@@ -122,52 +122,67 @@ func (m *JobStore) GetJob(ctx context.Context, id string) (*models.Job, error) {
...
+	}
+
```
rahulPalliyalil, line 182: (suggestion): extra space

rahulPalliyalil (approved):

api/jobs.go
justinpjose: I wrote the comments so that we can understand the high level of what the handler is doing by just reading the comments alone. And then if needed, we can dig through the code

mongo/jobs.go
justinpjose: Same as **COMMENT 1**

api/jobs.go
justinpjose: In this particular case, I think the extra space is nice to clearly separate the setting of the `options` and then calling the function which uses this `options` afterwards

mongo/jobs.go
justinpjose: I have removed this if statement entirely because it is unnecessary. Finding 0 or more jobs is completely fine and it should still be considered as successful

rahulPalliyalil (approved):

PR #113, opened 2022-05-25
add etag functionality for GET `/jobs/<job_id>/tasks/<task_id>`

justinpjose:
### What
- Ensure that we check the `If-Match` header in the request if it exists
- Compare ETag from request with ETag of the job resource to check for conflict
- Ensure we set `ETag` header in the response back
- Update unit test and component test

### How to review
- Check if the test passes

### Who can review
- Anyone

### DIFF
--- a/api/tasks_test.go
+++ b/api/tasks_test.go
@@ -44,7 +44,7 @@ var createTaskPayloadFmt = `{
 	"number_of_documents": 5
 }`
 
-func expectedTask(cfg *config.Config, jobID, taskName string, jsonResponse bool, lastUpdated time.Time, numberOfDocuments int) models.Task {
+func expectedTask(ctx context.Context, cfg *config.Config, t *testing.T, jobID, taskName string, jsonResponse bool, lastUpdated time.Time, numberOfDocuments int) models.Task {
 	task := models.Task{
 		JobID:       jobID,
 		LastUpdated: lastUpdated,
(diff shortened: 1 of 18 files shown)


api/tasks_test.go
```diff
@@ -59,12 +59,45 @@ func expectedTask(cfg *config.Config, jobID, taskName string, jsonResponse bool,
 	if jsonResponse {
 		task.Links.Job = fmt.Sprintf("%s/%s%s", cfg.BindAddr, cfg.LatestVersion, task.Links.Job)
 		task.Links.Self = fmt.Sprintf("%s/%s%s", cfg.BindAddr, cfg.LatestVersion, task.Links.Self)
+	} else {
+		taskETag, err := models.GenerateETagForTask(ctx, task)
+		if err != nil {
+			t.Errorf("failed to generate eTag for expected test job - error: %v", err)
```
cookel2, line 65: Why does the GenerateETagForTask function only get called if the jsonResponse is false? Perhaps a comment to explain this would be useful here. I notice that in the expectedJob function it gets called whatever the jsonResponse is: https://github.com/ONSdigital/dp-search-reindex-api/blob/feature/etag-get-jobs-tasks-id/api/jobs_test.go#:~:text=%3A%3D%20models.-,GenerateETagForJob,-(ctx%2C 

Also a small typ here:  
```suggestion
			t.Errorf("failed to generate eTag for expected test task - error: %v", err)
```

cookel2 (changes requested):

cookel2 (approved):
looks good and works locally for me

PR #114, opened 2022-05-25
add etag functionality for PUT `/jobs/<id>/tasks/number_of_tasks/<count>`

justinpjose:
### What
- Ensure ETag is generated when writing partial update to job resource
  - Here we would need to get the existing job resource
  - Check the current `etag` value matches the IfMatch header value (if it is set and not equal to `*`)
  - Then transform the data - update `number_of_tasks`
  - Then generate the ETag and add to transformed resource
  - Then store in database
- Ensure we set ETag header in the response
- Add unit test and component test

### How to review
- Check if all the test passes

### Who can review
- Anyone

### DIFF
--- a/api/jobs_test.go
+++ b/api/jobs_test.go
@@ -1198,16 +1198,6 @@ func TestPutNumTasksHandler(t *testing.T) {
 
 	Convey("Given a Search Reindex Job API that updates the number of tasks for specific jobs using their id as a key", t, func() {
 		jobStoreMock := &apiMock.DataStorerMock{
-			PutNumberOfTasksFunc: func(ctx context.Context, id string, count int) error {
-				switch id {
-				case validJobID2:
-					return nil
-				case validJobID3:
-					return errors.New("unexpected error updating the number of tasks")
-				default:
-					return mongo.ErrJobNotFound
-				}
-			},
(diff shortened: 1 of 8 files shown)


api/jobs_test.go
```diff
@@ -1219,19 +1209,108 @@ func TestPutNumTasksHandler(t *testing.T) {
...
+
+				Convey("And the etag of the response job resource should be returned via the ETag header", func() {
+					So(resp.Header().Get(dpresponse.ETagHeader), ShouldNotBeEmpty)
+				})
+			})
+		})
+
+		Convey("When a request is made where If-Match header is set to *", func() {
 			req := httptest.NewRequest("PUT", fmt.Sprintf("http://localhost:25700/jobs/%s/number_of_tasks/%s", validJobID2, validCount), nil)
+			headers.SetIfMatch(req, "*")
```
cookel2, line 1251: Please handle SetIfMatch error here (line 1251) and also for lines 1268, 1285, and 1301.

cookel2 (changes requested):

api/jobs_test.go
justinpjose: Done

cookel2 (approved):
looks good and works locally

PR #116, opened 2022-06-09
update swagger spec with `ETag` information

justinpjose:
### What
- Update swagger spec with `ETag` information for each endpoint (except for `/health`)
- Upgrade go version to 1.18.3
- Upgrade dependencies

### How to review
- Check if the changes make sense

### Who can review
-  Anyone


api/jobs.go
```diff
@@ -386,8 +386,7 @@ func (api *API) PutNumTasksHandler(w http.ResponseWriter, req *http.Request) {
 	// set eTag on ETag response header
 	dpresponse.SetETag(w, updatedETag)
 
-	w.Header().Set("Content-Type", "application/json")
-	w.WriteHeader(http.StatusOK)
+	w.WriteHeader(http.StatusNoContent)
```
rahulPalliyalil, line 389: (question): why do you need a no content status here ?

justinpjose: This is because status code is more explicit as no content is send in the response payload body

rahulPalliyalil: ok, makes sense.

rahulPalliyalil (approved):

PR #125, opened 2022-10-11
Add client sdk for number of docs api

RMPall:
### What

Add client sdk for number of docs api.

### How to review

Please check if the changes make sense.

### Who can review

Anyone except me.


justinpjose (commented):
This looks good. Unit test needs to be written for the client function

RMPall:
> This looks good. Unit test needs to be written for the client function
I think none of the client sdk has been unit tested, there is only one test for the etag header. This makes sense as all the sdks are just calling the api endpoints, so there is not much value that you can get from this.

RMPall:
> This looks good. Unit test needs to be written for the client function

Not sure if it adds any value here as the endpoints are clearly unit tested . This is just an sdk reaching out to the endpoint. Anyways the tests are added.

justinpjose:
> > This looks good. Unit test needs to be written for the client function
> 
> Not sure if it adds any value here as the endpoints are clearly unit tested . This is just an sdk reaching out to the endpoint. Anyways the tests are added.

it’s worth having good test coverage even on ones that don’t add much value

